PR #7, opened 2018-07-25
Fix various makefile issues with JDK11

ChengJin01:
The changes are to resolve all makefile issues
when compiling JDK11 on Windows.

Signed-off-by: CHENGJin <jincheng@ca.ibm.com>


closed/make/copy/Copy-java.base.gmk
```diff
@@ -29,6 +29,7 @@ ALT_JVMCFG_SRC := $(TOPDIR)/closed/src/java.base/conf/jvm.cfg
 override-jvm.cfg : $(LIB_DST_DIR)/jvm.cfg
 	$(ECHO) "Overriding jvm.cfg with J9VM version"
 	$(CP) $(ALT_JVMCFG_SRC) $(LIB_DST_DIR)/jvm.cfg
+	$(CP) $(ALT_JVMCFG_SRC) $(OUTPUTDIR)/jdk/lib/jvm.cfg
```
pshipton: Why is this required?

closed/OpenJ9.gmk
```diff
@@ -378,6 +378,12 @@ endif
 	@$(MKDIR) -p $(MODULES_LIBS_DIR)/java.base/server
 	@$(CP) -p $(OUTPUTDIR)/vm/redirector/$(LIBRARY_PREFIX)jvm_jdk11$(SHARED_LIBRARY_SUFFIX) $(MODULES_LIBS_DIR)/java.base/server/$(LIBRARY_PREFIX)jvm$(SHARED_LIBRARY_SUFFIX)
 
+ifeq (windows,$(OPENJDK_TARGET_OS))
+	@$(ECHO) "Updating vm/lib/jvm.* with vm/j9vm_jdk11/jdk11_jvm.*"
+	@$(CP) -p $(OUTPUTDIR)/vm/j9vm_jdk11/jdk11_jvm$(STATIC_LIBRARY_SUFFIX) $(OUTPUTDIR)/vm/lib/jvm$(STATIC_LIBRARY_SUFFIX)
+	@$(CP) -p $(OUTPUTDIR)/vm/j9vm_jdk11/jdk11_jvm.exp $(OUTPUTDIR)/vm/lib/jvm.exp
```
keithc-ca: These commands ought to use `$(LIBRARY_PREFIX)` even though it's normally empty.

pshipton: Does it still work without copying the .exp file?

ChengJin01: Agreed and added.

closed/src/java.base/conf/jvm.cfg
```diff
@@ -17,6 +17,8 @@
 # 
 # ===========================================================================
 -j9vm KNOWN
+-server IGNORE
```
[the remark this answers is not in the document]
ChengJin01: We still have the following setting in OpenJ9.gmk to deal with the /server directory 
e.g.
```
	@$(MKDIR) -p $(MODULES_LIBS_DIR)/java.base/server
	@$(CP) -p $(OUTPUTDIR)/vm/redirector/$(LIBRARY_PREFIX)jvm_jdk11$(SHARED_LIBRARY_SUFFIX) $(MODULES_LIBS_DIR)/java.base/server/$(LIBRARY_PREFIX)jvm$(SHARED_LIBRARY_SUFFIX)
```
If the /server should be totally removed, then all setting in jvm.cfg and Copy-java.base.gmk  won't be necessary.

pshipton:
@jdekonin do we have any PR builds for jdk11 yet?

jenkins compile xlinux

closed/src/java.base/conf/jvm.cfg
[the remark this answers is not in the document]
pshipton: Not sure why its different from jdk10, which doesn't IGNORE server.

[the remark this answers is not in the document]
pshipton: Although the file is there in the jdk10 extensions, perhaps its not used?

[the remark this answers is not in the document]
pshipton: We create a server directory intentionally, for compatibility with OpenJDK. So an app which looks in the server directory for the libjvm.so will find it and work. I don't want to remove that part, if that is what you meant.

pshipton:
@jdekonin Adam mentioned that only you can run jdk11 builds, however all committers should be able to start them from extensions, or from OpenJ9. They just shouldn't be enabled by default, but must be explicitly requested.

pshipton:
Anyway, I want to ensure these changes for Windows don't break the xLinux build.

closed/src/java.base/conf/jvm.cfg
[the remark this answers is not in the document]
ChengJin01: Yes. So all changes related to jvm.cfg should be kept there in favor of the operations on the server directory.

pshipton:
@ChengJin01 note that once we finalize these changes, they should also be delivered to the "head" stream https://github.com/ibmruntimes/openj9-openjdk-jdk

ChengJin01:
> once we finalize these changes, they should also be delivered to the "head" stream https://github.com/ibmruntimes/openj9-openjdk-jdk

If so,  I need to create a pull request on ibmruntimes/openj9-openjdk-jdk to verify the compilation result.

closed/OpenJ9.gmk
ChengJin01: Already recompiled and it seems there is no issue detected during compilation (also no problem with jdk/bin/java -version). So removed the setting for jvm.exp.

AdamBrousseau:
jenkins compile xlinux

closed/OpenJ9.gmk
keithc-ca: It should have been added to both arguments of `$(CP)`.

PR #9, opened 2018-08-01
Enable use of ccache

keithc-ca:
This is a replay of https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/11.


pshipton:
jenkins compile zlinux

PR #12, opened 2018-08-13
Fix compile warnings with gcc 7.3

keithc-ca:
* add to loop test so the compiler can see the index is valid
* avoid left shifts of negative values


keithc-ca:
Copyrights updated.

keithc-ca:
Updated so both big-endian and little-endian expressions are structurally similar as before (with missing casts added).

PR #13, opened 2018-08-23
Fix right shifts broken by #12

keithc-ca:
The right shift expressions were changed from `(-1) >> n` to `-(1 >> n)` which yields 0 when n>0.
This fixes that, and improves the expressions so they don't involve shifting signed values (it's implementation defined whether `(-1) >> 3` is -1 or -8.


keithc-ca:
See the discussion of right shift at https://en.wikipedia.org/wiki/Bitwise_operations_in_C where it notes
> most compilers will perform an arithmetic shift, causing the blank to be filled with the sign bit of the left operand

PR #28, opened 2018-10-02
Use new DDR ClassGenerator tool

keithc-ca:
* compile the java source against the generated classes
* merge ddr support makefiles & adjust names of intermediate targets

Requires https://github.com/eclipse/openj9/pull/3099


closed/make/DDR.gmk
```diff
@@ -0,0 +1,122 @@
...
+		openj9.traceformat \
+	)
+
+DDR_SRC_EXCLUDES := com/ibm/j9ddr/tools/ant
+
+ifeq (,$(filter mz31 mz64,$(OPENJ9_PLATFORM_CODE)))
+DDR_SRC_EXCLUDES += com/ibm/j9ddr/corereaders/tdump
+endif
+
+# Exclude source generated by legacy tool, if present.
```
pshipton: There is no legacy tool for Java 11, is this needed?

keithc-ca: Here, 'legacy' refers to the `PointerGenerator` and `StructureStubGenerator` tools that were in use prior to this change.

pshipton: I'm still not clear on where these legacy classes come from. Are they in git?

keithc-ca: Prior to this change, `DDR-gensrc.gmk` generated Java source into the source tree under `openj9/debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/structure` and `openj9/debugtools/DDR_VM/src/com/ibm/j9ddr/vm29/pointer/generated`; if any of those files are around, they need to be ignored.

pshipton:
@keithc-ca please take a look at the build failures

keithc-ca:
Updated for https://github.com/eclipse/openj9/pull/3190.

keithc-ca:
jenkins test all zlinux

closed/make/DDR.gmk
pshipton: So basically this handles re-building a pre-existing build with this change added?

keithc-ca: Yes.

pshipton:
This is causing a build failure on AIX https://github.com/eclipse/openj9/issues/3227, reverting until it can be fixed.

PR #38, opened 2018-10-12
Handle race condition in shared classes defineClass()

pdbain-ibm:
Handle IllegalArgumentException when parallel classloaders define the same
package. cf. eclipse/openj9#3038

Signed-off-by: Peter Bain <peter_bain@ca.ibm.com>

### DIFF
--- a/src/java.base/share/classes/java/net/URLClassLoader.java
+++ b/src/java.base/share/classes/java/net/URLClassLoader.java
@@ -1,6 +1,6 @@
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 1997, 2017 All Rights Reserved
+ * (c) Copyright IBM Corp. 1997, 2018 All Rights Reserved
  * ===========================================================================
  */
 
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/java/net/URLClassLoader.java
```diff
@@ -1,6 +1,6 @@
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 1997, 2017 All Rights Reserved
+ * (c) Copyright IBM Corp. 2000, 2018 All Rights Reserved
```
pshipton: Why the change in the start date?

pshipton:
jenkins test sanity zlinux

src/java.base/share/classes/java/net/URLClassLoader.java
pdbain-ibm: Copy/paste error.

pshipton:
PR testing is here https://ci.eclipse.org/openj9/view/Pull%20Requests/job/PullRequest-Sanity-JDK11-linux_390-64_cmprssptrs-OpenJDK11/10/

PR #39, opened 2018-10-17
Fix spelling of openj9.dtfj

keithc-ca:
Signed-off-by: Keith W. Campbell <keithc@ca.ibm.com>


pshipton:
I note using jextract can result in `Error. Native library j9jextract cannot be found. Please check your path.`. Hopefully this fixes the problem.

keithc-ca:
Yes, this should fix that problem.

PR #52, opened 2018-10-30
Replace Oracle libjsig.so with OpenJ9 version

andrew-m-leonard:
Signed-off-by: andrew-m-leonard <andrew_m_leonard@uk.ibm.com>

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -179,6 +179,17 @@ $(foreach subdir, j9vm server, \
 		$(subdir)/$(call SHARED_LIBRARY,jvm), \
 		redirector/$(call SHARED_LIBRARY,jvm_jdk11)))
 
+# jsig
+$(call openj9_add_jdk_special, \
+        server/$(call SHARED_LIBRARY,jsig), \
+        $(call SHARED_LIBRARY,jsig))
+$(call openj9_add_jdk_special, \
+        j9vm/$(call SHARED_LIBRARY,jsig), \
+        $(call SHARED_LIBRARY,jsig))
+$(call openj9_add_jdk_special, \
+        $(call SHARED_LIBRARY,jsig), \
+        $(call SHARED_LIBRARY,jsig))
+
 # java.base
 
 $(call openj9_add_jdk_shlibs, java.base, \
--- a/make/lib/Lib-java.base.gmk
+++ b/make/lib/Lib-java.base.gmk
@@ -23,6 +23,10 @@
 # questions.
 #
 
+# ===========================================================================
+# (c) Copyright IBM Corp. 2018, 2018 All Rights Reserved
+# ===========================================================================
+
 include LibCommon.gmk
 
 # Hook to include the corresponding custom file, if present.
@@ -126,6 +130,8 @@ endif
 ################################################################################
 # Create the jsig library
 
+# Don't build Oracle's libjsig for OpenJ9
+ifeq ($(OPENJ9_TOPDIR),)
 ifeq ($(OPENJDK_TARGET_OS_TYPE), unix)
   ifeq ($(STATIC_BUILD), false)
 
@@ -176,6 +182,7 @@ ifeq ($(OPENJDK_TARGET_OS_TYPE), unix)
 
   endif
 endif
+endif #ifeq ($(OPENJ9_TOPDIR),)
 
 ################################################################################
 # Create the symbols file for static builds.


andrew-m-leonard:
jenkins compile xlinux

make/lib/Lib-java.base.gmk
```diff
@@ -176,6 +182,7 @@ ifeq ($(OPENJDK_TARGET_OS_TYPE), unix)
 
   endif
 endif
+endif
```
keithc-ca: A comment to help match this with the opening `ifeq` would be helpful.

closed/OpenJ9.gmk
```diff
@@ -179,6 +179,14 @@ $(foreach subdir, j9vm server, \
 		$(subdir)/$(call SHARED_LIBRARY,jvm), \
 		redirector/$(call SHARED_LIBRARY,jvm_jdk11)))
 
+# jsig
+$(call openj9_add_jdk_special, \
+                server/$(call SHARED_LIBRARY,jsig), \
```
keithc-ca: This should add jsig to the `j9vm` subdirectory (not just the `server` subdirectory.

Please also respect the existing indentation style of this file.

closed/OpenJ9.gmk
```diff
@@ -179,6 +179,14 @@ $(foreach subdir, j9vm server, \
...
+                $(call SHARED_LIBRARY,jsig))
+$(call openj9_add_jdk_special, \
+                $(call SHARED_LIBRARY,jsig), \
```
keithc-ca: This is redundant and should be removed: jsig is already in the `compressedrefs` (or `default`) subdirectory which is searched before the main `lib` directory.

keithc-ca: The call on line 183 puts it in the server (and as requested in the j9vm) directories; the same places users will find libjvm.so.

andrew-m-leonard:
Still WIP, having trouble trying to get J9 libjsig.so into right place in image!!

andrew-m-leonard:
I've figured why now, it was because I was working on a more indirect patching of the Lib-java.base.gmk using a "post" extension, however it was causing this issue, hence i'm reverting that.
thanks

make/lib/Lib-java.base.gmk
andrew-m-leonard: added

closed/OpenJ9.gmk
andrew-m-leonard: indentation corrected
subdirs match where libjsig.so was before

andrew-m-leonard:
@keithc-ca @pshipton If you could review again please?

andrew-m-leonard:
jenkins compile xlinux

pshipton:
As @keithc-ca mentioned earlier, to be consistent the libjsig.so also needs to be available in the j9vm directory. This is consistent with having it available in the openjdk server directory (which is only created in openj9 builds for compatibility with openjdk).

Also note that jdk8 will require similar fixes to copy libjsig.so to the correct places.

pshipton:
How do these changes affect Windows, which has a different directory layout?

keithc-ca:
OpenJDK builds for Windows don't have jsig shared libraries (at least not in the latest hotspot builds for jdk8 and jdk11 from https://adoptopenjdk.net/).

andrew-m-leonard:
> 
> 
> As @keithc-ca mentioned earlier, to be consistent the libjsig.so also needs to be available in the j9vm directory. This is consistent with having it available in the server directory (which is only created in openj9 builds for compatibility with openjdk).
> 
> Also note that jdk8 will require similar fixes to copy libjsig.so to the correct places.

I've added it to j9vm.

jdk8 does not have an OpenJDK libjsig.so, so we don't have this issue.

andrew-m-leonard:
Ready again..

andrew-m-leonard:
jenkins compile xlinux

andrew-m-leonard:
jenkins test sanity xlinux

pshipton:
Do these changes affect Windows, which for an OpenJDK build doesn't contain jsig.dll? i.e. does it copy jsig.dll into the server, j9vm, lib directory on Windows?

> jdk8 does not have an OpenJDK libjsig.so, so we don't have this issue.

I believe there is a similar issue, in that the OpenJ9 libjsig.so should be available outside of the OpenJ9 compressedrefs/default directory, since user code must link with this library when using signals. Anyway I can open a new Issue for that.

pshipton:
On Windows the correct directories are bin\server and bin\j9vm and bin. Keith also pointed out the jsig.lib file is also required on Windows, although that could be handled via another PR.

keithc-ca:
Yes, let's defer the Windows issues to a separate PR. I don't think the changes here will negatively affect Windows.

pshipton:
We also need to consider AIX, which uses a libjsig.a file.

pshipton:
Ignore the previous comment, there is no libjsig.a file on AIX.

PR #58, opened 2018-11-02
Omit jsig shared library from compressedrefs/default directory

### LINKED ISSUE #57: libjsig.so / jsig.dll not required in vm directory
#52 copies the jsig library into the places where it should be. It doesn't need to be in the compressedrefs/default directory.

keithc-ca:
Fixes #57.


pshipton:
jenkins test sanity win

PR #73, opened 2018-12-05
OpenSSL crypto support for OpenJ9-OpenJDK 11 build

enasser:
Use the OpenSSL crypto algorithms with OpenJ9-OpenJDK for better performance.
Provide a switch in the existing security providers to switch to the native
crypto implementation instead of creating a new security provider.

Three crypto algorithms from OpenSSL are enabled for the native crypto. They
are Digest [only SHA algorithms], Cipher Block Chaining [CBC] and Galois Counter
Mode [GCM].


keithc-ca:
Several files appear to be copies of openjdk code (presumably modified). I suggest the changes should be made in place rather than moving to the `closed` folder so we will notice and benefit from any upstream changes to openjdk.

Some makefiles are modified rather than using the normal 'hook' mechanism.

andrew-m-leonard:
jenkins sanity xlinux

andrew-m-leonard:
jenkins compile xlinux

andrew-m-leonard:
jenkins test sanity xlinux

enasser:
> Several files appear to be copies of openjdk code (presumably modified). I suggest the changes should be made in place rather than moving to the closed folder so we will notice and benefit from any upstream changes to openjdk.

@keithc-ca For the openssl support, our design approach was to use a switch between the openssl crypto implementation and the default java crypto implementation instead of creating a separate crypto providers. Hence, the crypto driver programs (SunEntries & CipherCore) are modified to switch between the implementation based on system properties. The basic structure of the java interface programs for the openssl is created based on the existing default java crypto implementation programs . However, it need not be updated along with upstream changes of default crypto programs as the actual implementation of crypto is with openssl implementation. Also, keeping the openssl native interface call in the existing java crypto implementation programs make it difficult for the upstream merge of openjdk  due to merge conflicts. Hence, we decide to keep openssl java interface programs in the closed path.

pshipton:
> However, it need not be updated along with upstream changes of default crypto programs as the actual implementation of crypto is with openssl implementation.

This is only true if openssl is enabled. It means disabling openssl could fall back to an out of date crypto implementation.

> keeping the openssl native interface call in the existing java crypto implementation programs make it difficult for the upstream merge of openjdk due to merge conflicts

Isn't there only a merge conflict when something changes? In which case the change needs to be merged in for when openssl support is disabled.

enasser:
> This is only true if openssl is enabled. It means disabling openssl could fall back to an out of date crypto implementation.

@pshipton Only the openssl support related code is kept in the closed path. The existing java crypto implementation continues to be in the openjdk path and will be updated with upstream merge. Hence, the java crypto implementation will be always current and use the updated java crypto implementation when openssl is disabled.

> Isn't there only a merge conflict when something changes? In which case the change needs to be merged in for when openssl support is disabled.

I meant the openssl support code is kept in the closed path to avoid merge conflicts with the existing java crypto implementation. However, the existing java crypto implementation in the openjdk path will be updated during upstream merge.

enasser:
> Some of the modified Oracle GPL+CE files have a duplicate GPL license below the IBM Copyright. We should only have the IBM Copyright if there is already an Oracle GPLv2 + CE above it...

@andrew-m-leonard I have modified IBM Copyright as you advised. I was under the impression that we have to add "Classpath" exception in all files in the closed path. Hence, added the complete IBM Copyright statement.

> The following files should be implemented in closed/... make extensions files:
> make/lib/Lib-java.base.gmk
> make/copy/Copy-java.base.gmk

Done. Created the changes as a new revision. Will squash it once you complete the review.

andrew-m-leonard:
jenkins test sanity xlinux

pshipton:
jenkins test sanity xlinux

PR #74, opened 2018-12-06
Include j9aixbaddep in test-image artifacts

keithc-ca:
Fixes: eclipse/openj9#3221


DanHeidinga:
Jenkins test sanity aix jdk11

PR #83, opened 2018-12-18
Removing the dependency of libssl from libjncrypto

enasser:
The library dependency of libssl is not required for the library
libjncrypto as it is using only the functions from libcrypto.

This is related to eclipse/openj9/issues/3636

Signed-off-by: Nasser Ebrahim <enasser@in.ibm.com>


pshipton:
jenkins compile win,xlinux,osx jdk11

pshipton:
jenkins compile zlinux

pshipton:
Windows compile passed although jenkins shows it is still running.

PR #84, opened 2018-12-18
Use 64-bit MSVC toolchain when available

0dvictor:
64-bit toolchain is preferred because of compilation speed.

Signed-off-by: Victor Ding <dvictor@ca.ibm.com>


andrew-m-leonard:
This seems reasonable for contribution to openjdk, i'll raise a enhancement contribution

PR #95, opened 2019-01-21
Revert "Revert "Provide implementations of dladdr() with classpath ex…

andrew-m-leonard:
Should not have reverted the fix as the OpenJDK fix isn't in jdk-11.0.2+7, it is only 11.0.3...

This reverts commit 09a7d724717cdee6893d5990c1b5b9db3c83e82e.


andrew-m-leonard:
jenkins compile xlinux

andrew-m-leonard:
Copyright dates are correct for when change was done.

PR #97, opened 2019-01-22
Put jsig libraries in the correct places

### LINKED ISSUE #56: OpenJ9 builds do not include jsig.lib on Windows
The jsig.lib file should be included in the lib directory.

keithc-ca:
* on Windows the shared library should be in `bin`, `bin/j9vm` and `bin/server` (in an SDK, the static library should be in `lib`)
* on other platforms, the shared library should be in `lib`, `lib/j9vm` and `lib/server`
* also put the redirector static library where openjdk expects it so adjustments to `LDFLAGS_JDKLIB` are not necessary

This should fix #56.


pshipton:
jenkins test sanity win, xlinux

pshipton:
jenkins test sanity win,xlinux

PR #102, opened 2019-01-24
crypto library does not load on osx and windows jdk11

enasser:
The issue with mac OSX is that libjncrypto.dylib is linked to
/usr/local/lib/libcrypto.1.1.dylib instead of linking with rpath
i.e @rpath/libcrypto.1.1.dylib.
This is because libcrypto.1.1.dylib by default is built with
/usr/local/lib as the path instead of rpath.
To fix the issue, updated the path entry of libcrypto.1.1.dylib
with @rpath using install_name_tool.

Signed-off-by: Nasser Ebrahim <enasser@in.ibm.com>

PR #105, opened 2019-01-25
crypto library does not load on osx and windows jdk11

pshipton:
The issue with mac OSX is that libjncrypto.dylib is linked to
/usr/local/lib/libcrypto.1.1.dylib instead of linking with rpath
i.e @rpath/libcrypto.1.1.dylib.
This is because libcrypto.1.1.dylib by default is built with
/usr/local/lib as the path instead of rpath.
To fix the issue, updated the path entry of libcrypto.1.1.dylib
with @rpath using install_name_tool.

Deliver #102 to the openj9-0.12 branch

Signed-off-by: Nasser Ebrahim <enasser@in.ibm.com>


pshipton:
@keithc-ca

keithc-ca:
Confirmed the content is the same as already merged into openj9: merging.

Issue #104, opened 2019-01-24
Keep alive native resources


pshipton:
There is a potential problem in code like the following. It may be possible for the RSAPublicKeyImpl  to be finalized and the nativePtr freed while the native code is using it.

The solution is to use a `try{ } finally` to put a call to `java.lang.ref.Reference.reachabilityFence(key);` after the native call. Or for Java 8, `java.nio.Bits.keepAlive(key);`.

https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/0e39e88d9eb3ac437570398b736e96ff49714352/closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java#L103

pshipton:
@jerrylui803 @ashbm5

pshipton:
See the comments in #106, there are no changes required

PR #106, opened 2019-01-25
Keep RSA native resource alive

jerrylui803:
issue https://github.com/ibmruntimes/openj9-openjdk-jdk11/issues/104

Signed-off-by: Jerry Lui <jerrylui@ibm.com>

### DIFF
--- a/closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
+++ b/closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
@@ -99,8 +99,12 @@ public final class NativeRSACore {
 
         BigInteger n = key.getModulus();
         byte[] output = new byte[getByteLength(n)];
-
-        int outputLen = NativeCrypto.RSAEP(msg, msg.length, output, nativePtr);
+        int outputLen;
+        try {
+            outputLen = NativeCrypto.RSAEP(msg, msg.length, output, nativePtr);
+        } finally {
+            java.lang.ref.Reference.reachabilityFence(key);
+        }
 
         if (outputLen == -1) {
             return null;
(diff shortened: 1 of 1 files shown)


pshipton:
Looks good, although I expect the other algorithms using openssl need a similar change.

I changed the description so merging this doesn't close #104, because similar change are required for jdk8, even if the RSA code isn't delivered to jdk8 yet.

ashbm5:
@pshipton in all other OpenSSL backed crypto operations: Digest, AES-CBC and AES-GCM there isn't this kind of behaviour where a native routine takes in a ptr to a native resource that may be deallocated due to object finalization in Java. This is something unique so far for RSA due to the interaction between RSACore and RSA Key classes. In the other crypto operations there is only a single crypto class that takes in primitive types and a native pointer that is allocated and de-allocated by that same crypto class.

@jerrylui803 please confirm.

jerrylui803:
> @pshipton in all other OpenSSL backed crypto operations: Digest, AES-CBC and AES-GCM there isn't this kind of behaviour where a native routine takes in a ptr to a native resource that may be deallocated due to object finalization in Java. This is something unique so far for RSA due to the interaction between RSACore and RSA Key classes. In the other crypto operations there is only a single crypto class that takes in primitive types and a native pointer that is allocated and de-allocated by that same crypto class.
> 
> @jerrylui803 please confirm.

confirmed

pshipton:
Confirmed the following is sufficient to keep the key alive, so this change is not required.

https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/eaa2f0279376379fc0b47bb9a8fdeb0f236a80b6/src/java.base/share/classes/sun/security/rsa/RSACore.java#L128

https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/eaa2f0279376379fc0b47bb9a8fdeb0f236a80b6/src/java.base/share/classes/sun/security/rsa/RSACore.java#L157

PR #108, opened 2019-01-28
Fix compiler warnings when OpenSSL is enabled

keithc-ca:
Similar to https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/249, but for Java 11:
* fix casts between pointers and scalars (of differing sizes)
* fix assignment of const pointer to non-const pointer
* fix potentially uninitialized local variables


pshipton:
jenkins compile windows

pshipton:
jenkins compile win

PR #116, opened 2019-01-31
Add dependency in Main.mk to support the generation of jvmti.h

MarkQingGuo:
Add dependency in Main.mk to support the generation of jvmti.h

Generating jvmti.h before initialization of macros in Main.mk.

There are some additional information:
- The related issue is https://github.com/eclipse/openj9/issues/2368
- Add dependency in Main.mk to support the generation of jvmti.h before it starts to build openj9

Signed-off-by: MarkQingGuo <Qing.Guo@ibm.com>

### DIFF
--- a/closed/make/Main.gmk
+++ b/closed/make/Main.gmk
@@ -41,9 +41,9 @@ OPENSSL_MAKE := $(MAKE) -f $(TOPDIR)/closed/openssl.gmk SPEC=$(SPEC)
 openssl-build : buildtools-langtools
 	+$(OPENSSL_MAKE)
 
-java.base-copy : openssl-build
+java.base-copy : j9vm-build
 
-java.base-libs : java.base-copy j9vm-build
+java.base-libs : java.base-copy
 
 j9vm-build : openssl-build
 	+$(OPENJ9_MAKE) openj9_build_jdk
--- a/closed/make/copy/Copy-java.base.gmk
+++ b/closed/make/copy/Copy-java.base.gmk
@@ -23,7 +23,7 @@ TARGETS += \
 	$(INCLUDE_TARGET_DIR)/ibmjvmti.h \
 	#
 
-$(INCLUDE_TARGET_DIR)/jvmti.h : $(TOPDIR)/openj9/runtime/include/jvmti.h
+$(INCLUDE_TARGET_DIR)/jvmti.h : $(OUTPUTDIR)/vm/include/jvmti.h
 	$(call install-file)
 
 $(INCLUDE_TARGET_DIR)/ibmjvmti.h : $(TOPDIR)/openj9/runtime/include/ibmjvmti.h


tajila:
please add a link to the dependent PR in openj9

MarkQingGuo:
[https://github.com/eclipse/openj9/pull/4409](url)

tajila:
@keithc-ca Please take a look at this PR. 

Is this what you had in mind?

More details can be found here https://github.com/eclipse/openj9/pull/4409

closed/make/Main.gmk
```diff
@@ -27,6 +27,11 @@ CLEAN_DIRS += vm
 	test-image-openj9 \
 	#
 
+ifeq (,$(BUILD_JAVA_SPEC_HEADERS)) 
+  $(shell m4 -D JAVA_SPEC_VERSION=$(VERSION_FEATURE) $(OPENJ9_TOPDIR)/runtime/include/jvmti.h.m4 > $(OPENJ9_TOPDIR)/runtime/include/jvmti.h)
+  BUILD_JAVA_SPEC_HEADERS := "true"
+endif
```
keithc-ca: I believe this belongs in openj9 and should be in a make rule rather than executed as the makefile is read (which will at least trigger rebuilds of everything that depends on jvmti.h).

MarkQingGuo: jvmti.h is required for building target java.base.copy which is prior to targets in openj9. I tried to add a new target in openj9 but it seems doesn't work.

MarkQingGuo: I even tried to add the dependency at CMAKELIST.txt(before building target sourcetool) in the root directory in openj9, but it failed when built with Cmake. If it build by UMA, adding a target in GNUMakefile in openj9/runtime solves the issue but cmake is still a problem.

keithc-ca: You'll want to update `closed/make/copy/Copy-java.base.gmk`, adding a rule to run `m4` to generate `jvmti.h` (within `$(OUTPUTDIR)` somewhere) and adjust the rule
```
$(INCLUDE_TARGET_DIR)/jvmti.h : $(TOPDIR)/openj9/runtime/include/jvmti.h
```
so it depends instead on that file generated by `m4`.

MarkQingGuo: I added this dependency before, target java.base.copy is only one case. I tried to build it by UMA in my local machine, java.base.copy is built later. j9nonbuilder.h also needs jvmti.h and it is built before java.base.copy. So I add a same target in runtime/GNUmakefile, which means we have two parts for the same purpose. I also tried to build it on Jenkins, it seems it need jvmti.h before target build-openj9 or java.base. I added dependency to these two targets but no one works. I guess jvmti.h is needed just after configuration.

MarkQingGuo: That is not only one target needs jvmti.h, so we have to generate it early.

keithc-ca: OpenJDK generates jvmti.h so we should be able to do likewise.

keithc-ca: Perhaps what is needed is an adjustment to `closed/make./Main.gmk` (lines 46-48) to this:
```
java.base-libs : java.base-copy

j9vm-build : java.base-copy
```
This will force java.base-copy to build before j9vm-build. If java.base-copy ensures jvmti.h is generated and available, all should be good.

MarkQingGuo: I changed the rule to `java.base-libs : j9vm-build`, otherwise it won't compile
If the rule added depends on the file generated by m4, we have to create directory vm/include first.
I think we might have two methods here:
1. generate file in source code openj9/runtime/include, then remove it when clean it.
2. create a new directory, but it means we will create an existing directory later. But this part of code is in extension repo/make, I'm not sure if I have right to modify it.

closed/make/copy/Copy-java.base.gmk
```diff
@@ -35,7 +35,11 @@ TARGETS += \
 	$(INCLUDE_TARGET_DIR)/ibmjvmti.h \
 	#
 
-$(INCLUDE_TARGET_DIR)/jvmti.h : $(TOPDIR)/openj9/runtime/include/jvmti.h
+#$(OUTPUTDIR)/vm/include/jvmti.h : $(OPENJ9_TOPDIR)/runtime/include/jvmti.h.m4 
+#	m4 -D JAVA_SPEC_VERSION=$(VERSION_FEATURE) $< > $@
+
+$(INCLUDE_TARGET_DIR)/jvmti.h : $(TOPDIR)/openj9/runtime/include/jvmti.h.m4
+	m4 -D JAVA_SPEC_VERSION=$(VERSION_FEATURE) $< > $(TOPDIR)/openj9/runtime/include/jvmti.h
```
keithc-ca: I don't expect this will work: currently, openj9 source is copied into `$(OUTPUTDIR)/vm` and so writing into `$(TOPDIR)/openj9` will only help the _next_ build cycle. I think the rule (commented out) above is correct.

MarkQingGuo: I tried to build with the rule above before and let the build stop when building this target, it seems it will copy openj9 source later. When the above rule is completed, only `sourcetool` directory is copied into `$(OUTPUTDIR)`. The later rule will compile in my machine no matter build by Uma  or Cmake.

keithc-ca: The rule as written doesn't produce the target file and we don't want files written into the source tree.

The dependencies
`j9vm-build -> java.base-copy -> jvmti.h`
will ensure things are done in the right order (even though work is done in parallel).

MarkQingGuo:
@keithc-ca Hope this one is what in your mind. However, when building target `java.base-copy`, directory `$(OUTPUTDIR)/vm/` doesn't exist at this point, we may create `$(OUTPUTDIR)/vm/include` before copying openj9 source, otherwise, we have to create a same rule somewhere later.

keithc-ca:
It may be necessary to add `$(MKDIR) -p $(@D)` before the invocation of `m4` to ensure the target directory exists.

MarkQingGuo:
build successfully in Jenkins.

MarkQingGuo:
Jenkins sanity and functional test pass, but it's unstable with some skipped test

closed/make/copy/Copy-java.base.gmk
```diff
@@ -35,8 +35,9 @@ TARGETS += \
 	$(INCLUDE_TARGET_DIR)/ibmjvmti.h \
 	#
 
-$(INCLUDE_TARGET_DIR)/jvmti.h : $(TOPDIR)/openj9/runtime/include/jvmti.h
-	$(call install-file)
+$(INCLUDE_TARGET_DIR)/jvmti.h : $(TOPDIR)/openj9/runtime/include/jvmti.h.m4
+	$(MKDIR) -p $(OUTPUTDIR)/vm/include
+	m4 -D JAVA_SPEC_VERSION=$(VERSION_FEATURE) $< | tee -a $@ $(OUTPUTDIR)/vm/include/jvmti.h > /dev/null
```
keithc-ca: If the file in `$(OUTPUTDIR)/vm` is needed in addition to the target of this rule, split the rule into two to produce both, but each rule should normally create the target it claims to produce. Use `$(@D)` to refer to the directory containing the target file rather than repeating paths.

Please remove the use of `tee` - particularly the use of the `-a` option which will append to an existing file rather than create the desired content.

MarkQingGuo: Do you mean that split `$(INCLUDE_TARGET_DIR)/jvmti.h` and `$(OUTPUTDIR)/vm/include/jvmti.h`? If so, I have split them into two rules and test building successfully in my local machine.

closed/make/copy/Copy-java.base.gmk
```diff
@@ -36,8 +36,9 @@ TARGETS += \
 	#
 
 $(OUTPUTDIR)/vm/include/jvmti.h: $(TOPDIR)/openj9/runtime/include/jvmti.h.m4
-  $(MKDIR) -p $(@D)
+	$(MKDIR) -p $(@D)
 	m4 -D JAVA_SPEC_VERSION=$(VERSION_FEATURE) $< > $@
+	@$(ECHO) "END this PRO"
```
keithc-ca: This is looking good. I'm assuming you'll remove this echo command once you have verified it is all working.

MarkQingGuo: build successfully in Jenkins, link to it https://hyc-runtimes-jenkins.swg-devops.com/job/Build-JDK11-linux_x86-64_cmprssptrs/1392/

keithc-ca:
Before this can be merged, we need changes for jdk8, jdk12 and jdknext.

We'll also need a strategy that doesn't break IBM internal builds (due to the removal of jvmti.h in favour of jvmti.h.m4). For this, I restate my suggestion that the solution should be in the build process of openj9 itself.

keithc-ca:
Jenkins sanity.functional zlinux jdk11

keithc-ca:

PR #117, opened 2019-02-01
Turn off native digest

jerrylui803:
This change will disable OpenSSL based accelerated Digest, it cannot be re-enabled with the command line option. This was done due to a segfault under some conditions when the Digest is invoked. https://github.com/eclipse/openj9/issues/4530


pshipton:
There is nothing wrong with this change, but it is causing the OpenJ9 OpenSSL_0  and cmdLineTester_CryptoTest_0 testing to fail. The tests need to be fixed. Reverting until the fix is ready.

PR #122, opened 2019-02-01
Turn off native digest

pshipton:
Reverts ibmruntimes/openj9-openjdk-jdk11#120 to turn off native digest as per #117.


pshipton:
Tested in https://github.com/eclipse/openj9/pull/4555

PR #126, opened 2019-02-06
Stop adding j9jit29 when the target CPU is aarch64

knn-k:
This commit stops adding j9jit29 to java.base shlibs when the target
CPU is aarch64.
This is a temporary change until the JIT library for aarch64 becomes
available.

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -198,7 +198,6 @@ $(call openj9_add_jdk_shlibs, java.base, \
 	j9gc29 \
 	j9gcchk29 \
 	j9hookable29 \
-	j9jit29 \
 	j9jnichk29 \
 	j9jvmti29 \
 	j9prt29 \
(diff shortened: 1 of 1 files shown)


knn-k:
Build of `libj9jit29.so` is blocked by the change in `runtime/compiler/module.xml` in eclipse/openj9#4447.

pshipton:
jenkins compile zlinux

keithc-ca:
I'd prefer to see a workaround in openj9 to produce a (stub?) j9jit. Have you explored that approach?

knn-k:
Development of aarch64 JIT is ongoing, but it will take a while.
I think removing j9jit29 from the aarch64 build is the best way to make the aarch64 runtime (interpreter only) automatically buildable, because this approach changes only one file and it can be reverted easily when the aarch64 JIT is ready.

keithc-ca:
I had hoped that having a stub JIT would be a smaller request, but perhaps I was just being naive.

Removing the JIT like this is likely to cause a number of tests to fail due the warning message about a missing JIT. Please confirm you're ok with that and I will merge this.

knn-k:
I accept the failures that would be caused by missing JIT compiler.
I would like this to be merged.  Thanks.

PR #130, opened 2019-02-08
Add launcher definition for jps

pdbain-ibm:
Requires https://github.com/eclipse/openj9/pull/4650

Signed-off-by: Peter Bain <peter_bain@ca.ibm.com>


keithc-ca:
I expect you'll need to remove the filtering of `jps` from `closed/make/launcher/Launcher-jdk.jcmd-post.gmk`.

pdbain-ibm:
> I expect you'll need to remove the filtering of `jps` from `closed/make/launcher/Launcher-jdk.jcmd-post.gmk`.

Empirically, that's not the case. The build system creates `jps` even with the exclusion.

PR #136, opened 2019-02-14
Stop overriding jvm.cfg with OpenJ9 version

keithc-ca:
* the file produced by openjdk works just fine

Fixes #131.

PR #138, opened 2019-02-15
openssl: ensure target directory exists before copying shared library

keithc-ca:
Previously was relying on the rule that would override jvm.cfg (which was removed by #136).


keithc-ca:
Jenkins compile win jdk11

Issue #142, opened 2019-03-19
Questions for jdk version infos


gaolibj:
We build a Java11 on IBM i platform.  Without adding any specify options related to jdk versions in configuration phase.   Following is the output of "java -version".

bash-4.4$ bin/java -version
openjdk version **"11.0.1-internal" 2018-10-16**
OpenJDK Runtime Environment (build **11.0.1**-internal+0-adhoc.gaolibj.openj9-openjdk-jdk11)
Eclipse OpenJ9 VM (build openj9-0.11.0, JRE 11 OS/400 ppc64-64-Bit Compressed References 20190311_000000 (JIT enabled, AOT enabled)
OpenJ9   - 090ff9d
OMR      - ea548a6
JCL      - d445507 based on **jdk-11.0.3+2**)

Get source codes with following steps.
-   git clone https://github.com/ibmruntimes/openj9-openjdk-jdk11.git
-   cd openj9-openjdk-jdk11
-   git checkout openj9-0.11
-   get_source.sh

The string **"11.0.1" 2018-10-16** is hardcoded in file openj9-openjdk-jdk11/make/autoconf/version-numbers as below.
```
DEFAULT_VERSION_FEATURE=11
DEFAULT_VERSION_INTERIM=0
DEFAULT_VERSION_UPDATE=1
DEFAULT_VERSION_PATCH=0
DEFAULT_VERSION_DATE=2018-10-16
```

The string **jdk-11.0.3+2** is set with OPENJDK_TAG in file
 openj9-openjdk-jdk11/closed/autoconf/custom-hook.m4.

```
AC_DEFUN_ONCE([OPENJDK_VERSION_DETAILS],
[
  OPENJDK_SHA=`git -C $TOPDIR rev-parse --short HEAD`
  LAST_TAGGED_SHA=`git -C $TOPDIR rev-list --tags="jdk-11*" --topo-order --max-count=1 2>/dev/null`
  if test "x$LAST_TAGGED_SHA" != x; then
    OPENJDK_TAG=`git -C $TOPDIR describe --tags "$LAST_TAGGED_SHA"`
  else
    OPENJDK_TAG=
  fi
  AC_SUBST(OPENJDK_SHA)
  AC_SUBST(OPENJDK_TAG)

  # Outer [ ] to quote m4.
  [ USERNAME=`$ECHO "$USER" | $TR -d -c '[a-z][A-Z][0-9]'` ]
[...]

PR #143, opened 2019-03-22
Improve git tag resolution

keithc-ca:
Don't consider tags attached to descendant commits.

See #142.


pshipton:
jenkins compile win,plinux,aix,osx

PR #144, opened 2019-03-25
Remove use of 'sort -V'

keithc-ca:
Use only options that are available on all platforms.

See: https://github.com/eclipse/openj9/issues/4932


pshipton:
jenkins compile plinux,win,aix,osx

PR #149, opened 2019-04-02
AArch64: Allow both native build and cross build

knn-k:
This commit changes custom-hook.m4.  It adds "_cross" to OPENJ9_BUILDSPEC
when OPENJ9_CPU is aarch64 and it is in a cross build environment.


knn-k:
https://github.com/eclipse/openj9/pull/5321 adds new spec files for AArch64 Linux with `_cross` in their names.

PR #151, opened 2019-04-04
Backporting OpenJDK fix JDK-8220782

adamfarley:
Defined as the fix for "Program fails when using JDK addressed by
UNC path and using Security Manager"

This will be contributed to OpenJDK shortly.

Signed-off-by: Adam Farley <adam.farley@uk.ibm.com>


adamfarley:
The PolicyFileTester testcase (proposed by the originator at JDK-8218618) has been run to prove this works, and I've also run the PolicyFile jtreg tests. 2 failures on both fixed and unfixed jres, so we can be somewhat sure this fix doesn't worsen the situation.

PR #152, opened 2019-04-04
Add a check for illegal key size for CBC and GCM with NativeCrypto

jerrylui803:
Signed-off-by: Jerry Lui <jerrylui@ibm.com>

### DIFF
--- a/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
+++ b/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
@@ -35,6 +35,7 @@ import java.io.*;
 import java.security.*;
 import javax.crypto.*;
 import static com.sun.crypto.provider.AESConstants.AES_BLOCK_SIZE;
+import com.sun.crypto.provider.AESCrypt;
 
 import jdk.crypto.jniprovider.NativeCrypto;
 
@@ -219,6 +220,11 @@ final class NativeGaloisCounterMode extends FeedbackCipher {
               throws InvalidKeyException {
         if (keyValue == null || ivValue == null) {
             throw new InvalidKeyException("Internal error");
+        }
+
+        if (!AESCrypt.isKeySizeValid(keyValue.length)) {
+            throw new InvalidKeyException("Invalid AES key length: " +
+                keyValue.length + " bytes");
         }
 
         this.key = keyValue.clone();
@@ -413,7 +419,7 @@ final class NativeGaloisCounterMode extends FeedbackCipher {
         if (len < tagLenBytes) {
             throw new AEADBadTagException("Input too short - need tag");
         }
- 
+
         // do this check here can also catch the potential integer overflow
         // scenario for the subsequent output buffer capacity check.
         checkDataLength(ibuffer.size(), (len - tagLenBytes));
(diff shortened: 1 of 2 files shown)


jerrylui803:
Here are 2 test files that shows a crash without this check: 
For CBC: 
```
import java.security.spec.AlgorithmParameterSpec;
import java.util.Random;
import javax.crypto.Cipher;
import javax.crypto.spec.GCMParameterSpec;
import javax.crypto.spec.SecretKeySpec;

public class CBCKeyTest {
    public static void main(String[] args) {
        final int len = 1024;
        final int tagLen = 16;
        // 15 bytes is an illegal key size
        final int key_size = 15;
        byte[] skey_bytes = new byte[key_size/8];
        SecretKeySpec skey = new SecretKeySpec(skey_bytes, "AES");
        byte[] data = new byte[len];
        byte[] tagBuffer = new byte[len + tagLen];
        byte[] iv = new byte[12];

        try {
            Cipher cipher = Cipher.getInstance("AES/CBC/NoPadding", "SunJCE");
            cipher.init(Cipher.ENCRYPT_MODE, skey);
            cipher.doFinal(data, 0, data.length, tagBuffer);

        } catch (Exception e) {
            e.printStackTrace();
            System.exit(1);
        }
    }
}
```

For GCM: 
```
import java.security.spec.AlgorithmParameterSpec;
import java.util.Random;
import javax.crypto.Cipher;
import javax.crypto.spec.GCMParameterSpec;
import javax.crypto.spec.SecretKeySpec;

public class GCMKeyTest {
    public static void main(String[] args) {
        final int len = 1024;
        final int tagLen = 16;
        // 15 bytes is an illegal key size
        final int key_size = 15;
        byte[] skey_bytes = new byte[key_size/8];
        SecretKeySpec skey = new SecretKeySpec(skey_bytes, "AES");
        byte[] data = new byte[len];
        byte[] tagBuffer = new byte[len + tagLen];
        byte[] iv = new byte[12];

        try {
            Cipher cipher = Cipher.getInstance("AES/GCM/NoPadding", "SunJCE");
            AlgorithmParameterSpec iva = new GCMParameterSpec(tagLen * 8, iv);
            cipher.init(Cipher.ENCRYPT_MODE, skey, iva);
            cipher.doFinal(data, 0, data.length, tagBuffer);

        } catch (Exception e) {
            e.printStackTrace();
            System.exit(1);
        }
    }
}
```

jerrylui803:
@ashbm5  @pshipton @keithc-ca

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
```diff
@@ -97,6 +99,16 @@
         nativeCrypto = NativeCrypto.getNativeCrypto();
     }
 
+    // check if the specified length (in bytes) is a valid keysize for AES
+    static final boolean isKeySizeValid(int len) {
```
keithc-ca: Does it make sense to reuse the implementation from `NativeCipherBlockChaining`?
If so, you might also refactor this so instead of a boolean function, we have one that throws `InvalidKeyException` if appropriate.

pshipton:
@jerrylui803 can you please add the test cases to the OpenJ9 openssl test suite to avoid future regressions.
https://github.com/eclipse/openj9/tree/master/test/functional/cmdLineTests/openssl

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeGaloisCounterMode.java
jerrylui803: just realized I can use `isKeySizeValid()` from AESCrypt.java directly, so I changed it to that.

pshipton:
Note testing is running in https://github.com/eclipse/openj9/pull/5404

PR #161, opened 2019-04-22
Add support for windows CMake builds

dnakamura:

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -454,37 +454,41 @@ ifneq (,$(OPENJ9_DEVELOPER_DIR))
 endif
 
 ifeq (true,$(OPENJ9_ENABLE_CMAKE))
-CMAKE_ARGS :=
+  CMAKE_ARGS :=
+  ifeq (windows,$(OPENJDK_TARGET_OS))
+    CMAKE_ARGS += "-DCMAKE_TOOLCHAIN_FILE=$(OUTPUTDIR)/toolchain-win.cmake"
+  else
+    # We grab the C/C++ compilers detected by autoconf or provided by user, not
+    # the CC/CXX variables defined by the makefiles, which potentially include
+    # the ccache command which will throw off cmake.
+    ifneq (,$(OPENJ9_CC))
+      CMAKE_ARGS += "-DCMAKE_C_COMPILER=$(OPENJ9_CC)"
+    else
+      CMAKE_ARGS += "-DCMAKE_C_COMPILER=$(ac_cv_prog_CC)"
+    endif
 
-# We grab the C/C++ compilers detected by autoconf or provided by user, not
-# the CC/CXX variables defined by the makefiles, which potentially include
-# the ccache command which will throw off cmake.
-ifneq (,$(OPENJ9_CC))
-  CMAKE_ARGS += "-DCMAKE_C_COMPILER=$(OPENJ9_CC)"
-else
-  CMAKE_ARGS += "-DCMAKE_C_COMPILER=$(ac_cv_prog_CC)"
-endif
+    ifneq (,$(OPENJ9_CXX))
+      CMAKE_ARGS += "-DCMAKE_CXX_COMPILER=$(OPENJ9_CXX)"
+    else
+      CMAKE_ARGS += "-DCMAKE_CXX_COMPILER=$(ac_cv_prog_CXX)"
+    endif
+  endif # ifeq (windows,$(OPENJDK_TARGET_OS))
 
-ifneq (,$(OPENJ9_CXX))
-  CMAKE_ARGS += "-DCMAKE_CXX_COMPILER=$(OPENJ9_CXX)"
-else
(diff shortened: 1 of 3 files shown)


closed/autoconf/custom-hook.m4
```diff
@@ -405,6 +405,13 @@ AC_DEFUN_ONCE([CUSTOM_LATE_HOOK],
 
   # explicitly disable classlist generation
   ENABLE_GENERATE_CLASSLIST=false
+
+  if test "x$OPENJDK_BUILD_OS" = xwindows ; then
+    OPENJ9_TOOL_DIR="$OUTPUTDIR/tools"
+    AC_SUBST([OPENJ9_TOOL_DIR])
+    OPENJ9_GENERATE_TOOL_WRAPPERS
```
pshipton: I don't see anything making this specific to cmake builds.

dnakamura: I can change it if you want. I figured it really didn't matter if we generate a few text files that potentially go unused

pshipton:
jenkins test sanity win

closed/autoconf/custom-hook.m4
```diff
@@ -540,3 +547,36 @@ AC_DEFUN([CONFIGURE_OPENSSL],
   AC_SUBST(BUILD_OPENSSL)
   AC_SUBST(OPENSSL_CFLAGS)
 ])
+
+# Create a tool wrapper for use by cmake.
+# Consists of a shell script which wraps commands with an invocation of fixpath.
```
keithc-ca: Why do we need this? Why can't we simply ask cmake to use `CC=fixpath cl ...`?

dnakamura: It appears to work the first time, but when rebuilding,  sometimes it just stops working. I believe it occurs when cmake decides it needs to re-run itself and needs to re-test the compilers.

keithc-ca: Wouldn't it be better (and perhaps simpler) to just fix that?

dnakamura: no. It is a limitiation/bug in cmake itself

closed/OpenJ9.gmk
```diff
@@ -454,22 +454,26 @@ ifneq (,$(OPENJ9_DEVELOPER_DIR))
...
+    else
+      CMAKE_ARGS += "-DCMAKE_C_COMPILER=$(ac_cv_prog_CC)"
+    endif
+
+    ifneq (,$(OPENJ9_CXX))
+      CMAKE_ARGS += "-DCMAKE_CXX_COMPILER=$(OPENJ9_CXX)"
+    else
+      CMAKE_ARGS += "-DCMAKE_CXX_COMPILER=$(ac_cv_prog_CXX)"
+    endif
+endif # OPENJ9_ENABLE_CMAKE
```
keithc-ca: This `endif` is for Windows, not CMake; and is not properly indented.

closed/autoconf/custom-hook.m4
keithc-ca: I'm not convinced: I think we should explore this further (later).

closed/autoconf/custom-hook.m4
```diff
@@ -579,3 +586,35 @@ AC_DEFUN([CONFIGURE_OPENSSL],
...
+# OPENJ9_GENERATE_TOOL_WRAPER(<name_of_wrapper>, <command_to_call>)
+AC_DEFUN([OPENJ9_GENERATE_TOOL_WRAPPER],
+[
+  tool_file="$OPENJ9_TOOL_DIR/$1"
+
+  echo "#!/bin/sh" > $tool_file
+  printf '%s "%s" "$[]@"\n' "$FIXPATH" "$2" >> $tool_file
```
keithc-ca: I think this deserves a comment that explains the use of `[]` in the format string.

closed/autoconf/custom-hook.m4
```diff
@@ -579,3 +586,37 @@ AC_DEFUN([CONFIGURE_OPENSSL],
...
+# Consists of a shell script which wraps commands with an invocation of fixpath.
+# OPENJ9_GENERATE_TOOL_WRAPER(<name_of_wrapper>, <command_to_call>)
+AC_DEFUN([OPENJ9_GENERATE_TOOL_WRAPPER],
+[
+  tool_file="$OPENJ9_TOOL_DIR/$1"
+
+  echo "#!/bin/sh" > $tool_file
+  # We need to insert an empty string ([]), to stop M4 treating "$@" as a
+  # variable reference
+  printf '%s "%s" "$[]@"\n' "$FIXPATH" "$2" >> $tool_file
```
keithc-ca: It's actually `/bin/sh` that we're talking to here, and now that the control string is enclosed in single-quotes, I don't think `[]` is necessary.

keithc-ca:
Jenkins test sanity win,win32,zlinux jdk11

PR #163, opened 2019-04-24
Verify the minimum NASM version requirement is met on x86-64

nbhuiyan:
OpenJ9 JIT requires NASM version 2.13+ on x86 platform.

Issue: eclipse/openj9#4653

Signed-off-by: Nazim Uddin Bhuiyan <Nazim.Uddin.Bhuiyan@ibm.com>

PR #165, opened 2019-04-29
Downgrade NASM requirement to version 2.11 on x86

nbhuiyan:
This is done to continue supporting building using NASM v2.11. #163 Set the requirement to NASM 2.13+, which is not necessary as AVX512 instructions are not being used at this time.

Signed-off-by: Nazim Uddin Bhuiyan <Nazim.Uddin.Bhuiyan@ibm.com>


nbhuiyan:
@keithc-ca Once you've reviewed this, I will open corresponding PRs for jdk8, 12 and 13 so that they can be merged at the same time.

PR #167, opened 2019-05-08
Native ChaCha20-Poly1305 implementation with OpenSSL

keithc-ca:
This extends #164 with a minor cleanup; its description is repeated below.

Native ChaCha20-Poly1305 and ChaCha20 implementation with OpenSSL.
Added NativeChaCha20Cipher.java and modified relevant classes to leverage OpenSSL by default.

Need OpenSSL 1.1.0 or above to enable Native ChaCha20 support.

There is an option jdk.nativeChaCha20=false to disable. The existing option to manipulate native crypto is also supported (jdk.nativeCrypto).


closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
```diff
@@ -68,13 +67,12 @@ private NativeCrypto() {
...
     @CallerSensitive
     public static NativeCrypto getNativeCrypto() {
-
         ClassLoader callerClassLoader = Reflection.getCallerClass().getClassLoader();
 
-        if ((callerClassLoader != null) && (callerClassLoader != VM.getVMLangAccess().getExtClassLoader())) {
-            throw new SecurityException("NativeCrypto");
-        }
-        return new NativeCrypto();
+        if ((callerClassLoader == null) || (callerClassLoader == VM.getVMLangAccess().getExtClassLoader()))
```
pshipton: Using `{ }` is preferred, to avoid future mistakes when adding more content to the body of the if statement.

pshipton:
jenkins compile aix,osx,win,plinux

pshipton:
jenkins compile aix,osx,win,plinux jdk11

closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
keithc-ca: Fixed.

pshipton:
jenkins test sanity osx jdk11

pshipton:
This is ready to merge, but I think it should be ported to the jdk stream first (or we'll have to track making sure it gets done later). @keithc-ca are you willing to check how difficult the port will be?

keithc-ca:
> @keithc-ca are you willing to check how difficult the port will be?

Sure, I'll have a look - I don't expect it will be too onerous.

keithc-ca:
Latest push includes some further cleanup discovered while making changes for jdk13:
* remove unused field
* don't fetch system property needlessly

pshipton:
jenkins compile plinux jdk11

PR #169, opened 2019-05-13
Rework static initialization of useNativeChaCha20Cipher

keithc-ca:
* use a helper method so the field can be final


keithc-ca:
The initialization code is now identical to that proposed in https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/107.

pshipton:
jenkins compile plinux jdk11

PR #170, opened 2019-05-14
Create thread in SystemThreadGroup and not inherit ThreadLocals

JasonFengJ9:
Create thread in `SystemThreadGroup` and not inherit `ThreadLocals`

Created the thread via `com.ibm.oti.vm.VM.getVMLangAccess().createThread()`.
This thread is expected to be alive along with `JVM` hence belongs to the system thread group.
This thread doesn't need inherit `ThreadLocals` and avoid copying `ThreadLocals` which causes `NPE` in some use scenarios.

Verified that this fixes https://github.com/eclipse/openj9/issues/5726.

Depends on: https://github.com/eclipse/openj9/pull/5778
issue: https://github.com/eclipse/openj9/issues/5726

Note: will create PRs for `Java 8/12/13` if this is approved.

Reviewer: @pshipton 
FYI: @DanHeidinga 

### DIFF
--- a/closed/src/java.base/share/classes/java/io/ClassCache.java
+++ b/closed/src/java.base/share/classes/java/io/ClassCache.java
@@ -1,6 +1,6 @@
 /*
 * ===========================================================================
-* (c) Copyright IBM Corp. 2017, 2018 All Rights Reserved
+* (c) Copyright IBM Corp. 2017, 2019 All Rights Reserved
 * ===========================================================================
  * 
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


src/java.base/share/classes/java/io/ObjectInputStream.java
```diff
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018,2018 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018,2019 All Rights Reserved
```
pshipton: Pls add the missing space, it should be `2018, 2019`

JasonFengJ9: Fixed.

pshipton:
How about we add a helper to java.lang.VMAccess so we can create the Reaper Thread without inheriting thread locals. I'm thinking the Thread should also be created in the system ThreadGroup rather than the current one.

pshipton:
Note that from Java 11 there is already a public API to create a Thread without inheriting the thread locals.

JasonFengJ9:
@pshipton this PR has been updated to create the thread in question via a new `VMLangAccess` method which belongs the system thread group and not inherit `ThreadLocals`. This PR depends on https://github.com/eclipse/openj9/pull/5778.
Please have another look.

pshipton:
This shouldn't close eclipse/openj9#5726 because we still need to fix jdk8 and other versions. I've updated the description.

closed/src/java.base/share/classes/java/io/ClassCache.java
```diff
@@ -209,25 +209,24 @@ private Object createEntry(CacheKey key) {
         }
 
         public Thread run() {
-            return new Reaper(cache, queue);
+            Reaper reaper = new Reaper(cache, queue);
+            return com.ibm.oti.vm.VM.getVMLangAccess().createThread(reaper, "ClassCache Reaper", true, false, true, null);
```
pshipton: This needs a doPrivileged()

pshipton: Hmm, actually maybe only for jdk8

pshipton: Please try with a SecurityManager installed

JasonFengJ9: There is a `doPrivileged()` from caller.
https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/771d7cb5fb5ee8d83a20bfc5646b3270f55b69af/closed/src/java.base/share/classes/java/io/ClassCache.java#L67-L68

JasonFengJ9: In addition, run a small program with `new ObjectInputStream(new FileInputStream("test.txt"))` with security manager enabled, no `AccessControlException` thrown (only requires access to the file being tested).
The code snippet should create `ClassCache` in question. 

P.S. the original Renaissance Suite requires more permissions to proceed hence use the small app above instead.

pshipton:
Please create a PR against the v0.14.2 branch.

PR #177, opened 2019-05-16
Resolve unused-result warning for jspawnhelper.c

andrew-m-leonard:
Signed-off-by: andrew-m-leonard <andrew_m_leonard@uk.ibm.com>

### DIFF
--- a/src/java.base/unix/native/jspawnhelper/jspawnhelper.c
+++ b/src/java.base/unix/native/jspawnhelper/jspawnhelper.c
@@ -49,7 +49,10 @@ extern int errno;
 #define ERR_ARGS 3
 
 void error (int fd, int err) {
-    write (fd, &err, sizeof(err));
+    if (write (fd, &err, sizeof(err)) != sizeof(err)) {
+        /* Not sure what to do here. I have no one to speak to. */
+        exit(0x80 + err);
+    }
     exit (1);
 }
 


andrew-m-leonard:
jenkins compile zlinux

pshipton:
I thought we were going to fix the code rather than disable the warning?

andrew-m-leonard:
I'm still a bit undecided, so see this as wip.. So I did some investigation of existing issues like this in openjdk, and the resolution in awt and core modules has been to disable the warning.
A code fix would be something a little weird like this:
if (write(..)) {
   // Ignore result...
}
I'm trying to do some testing at the moment, but struggling to find zLinux box with gcc7.4, seems the one that did have it now has gcc5.4 !?! ...

andrew-m-leonard:
So did a test build for jdkhead and it didn't fail... because it has been fixed in openjdk by:
https://bugs.openjdk.java.net/browse/JDK-8213017
I have requested a backport to jdk11u.
I will backport the fix to extensions until we get the backport.

andrew-m-leonard:
jenkins compile xlinux

pshipton:
jenkins compile plinux,zlinux jdk11

pshipton:
Shouldn't this be delivered to the openj9-staging branch? Will it cause a merge conflict if merged to openj9?

andrew-m-leonard:
I was thinking about that, I don't think it matters from the merge perspective as jspawnhelper.c has not been changed in either branch... but it does from the perspective that we want to test with openj9-staging (jdk-11.0.4+2) that has the issue then merge that all into openj9.....
So I will move it...

andrew-m-leonard:
jenkins compile plinux,zlinux jdk11

src/java.base/unix/native/jspawnhelper/jspawnhelper.c
```diff
@@ -22,6 +22,11 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2019, 2019 All Rights Reserved
```
pshipton: Should this have an IBM copyright if it was copied from the OpenJDK code?

pshipton:
> but it does from the perspective that we want to test with openj9-staging (jdk-11.0.4+2) that has the issue then merge that all into openj9

Agreed, but the acceptance build will verify this. If there is no merge conflict, the worst that would happen is the acceptance build continues to fail and we try again to fix it.

src/java.base/unix/native/jspawnhelper/jspawnhelper.c
andrew-m-leonard: not entirely sure... it's the way we've done it before!

andrew-m-leonard: the copyright checker will complain... as it's a change...

pshipton: Please check it. It doesn't seem correct to me to add IBM copyright on code that IBM didn't author.

andrew-m-leonard: hi, i'll remove, as you say it's then effectively identical to openjdk

andrew-m-leonard:
jenkins compile plinux,zlinux jdk11

andrew-m-leonard:
jenkins compile zlinux

PR #179, opened 2019-05-27
Associate *TraceFormat.dat files with the module openj9.traceformat

keithc-ca:
Issue: eclipse/openj9#4488.


pshipton:
jenkins test sanity xlinux jdk11

PR #181, opened 2019-05-29
Export OPENJDK_VERSION_NUMBER_FOUR_POSITIONS only for UMA

keithc-ca:
With eclipse/openj9#5838 `OPENJDK_VERSION_NUMBER_FOUR_POSITIONS` and `VERSION_MAJOR` don't need to be exported as widely.


pshipton:
jenkins compile win jdk11

PR #185, opened 2019-06-05
Generate DDR blob on cmake builds

dnakamura:

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -228,11 +228,16 @@ ifeq (true,$(OPENJ9_ENABLE_DDR))
 $(call openj9_add_jdk_bin, java.base, j9ddr.dat)
 
 .PHONY : run-ddrgen
-$(OUTPUTDIR)/vm/j9ddr.dat : run-ddrgen
+$(OPENJ9_VM_BUILD_DIR)/j9ddr.dat : run-ddrgen
+
 run-ddrgen :
+ifeq (true,$(OPENJ9_ENABLE_CMAKE))
+	$(MAKE) -C $(OPENJ9_VM_BUILD_DIR) j9ddr
+else
 	export CC="$(CC)" CXX="$(CXX)" $(EXPORT_MSVS_ENV_VARS) \
 		&& $(MAKE) -C $(OUTPUTDIR)/vm/ddr -f run_omrddrgen.mk
-endif
+endif # OPENJ9_ENABLE_CMAKE
+endif # OPENJ9_ENABLE_DDR
 
 $(eval $(call openj9_add_jdk_rules, \
 	$(call FindLibDirForModule, java.base)/openj9-notices.html, \
@@ -491,6 +496,7 @@ ifeq (true,$(OPENJ9_ENABLE_CMAKE))
     CMAKE_ARGS += -DCMAKE_C_COMPILER_LAUNCHER="$(ESCAPED_CCACHE)"
   endif # CCACHE
 
+  CMAKE_ARGS += -DOMR_DDR=$(OPENJ9_ENABLE_DDR)
   CMAKE_ARGS += $(EXTRA_CMAKE_ARGS)
 
 $(OUTPUTDIR)/vm/cmake.stamp :


closed/OpenJ9.gmk
```diff
@@ -231,11 +231,16 @@ ifeq (true,$(OPENJ9_ENABLE_DDR))
 $(call openj9_add_jdk_bin, java.base, j9ddr.dat)
 
 .PHONY : run-ddrgen
-$(OUTPUTDIR)/vm/j9ddr.dat : run-ddrgen
+$(OPENJ9_VM_BUILD_DIR)/j9ddr.dat : run-ddrgen
+ifeq (true,$(OPENJ9_ENABLE_CMAKE))
+run-ddrgen:
```
keithc-ca: Please move the conditional into the rule rather than repeating it.

dnakamura:
Marked as WIP as I realized that i depends on omr and openj9 changes

dnakamura:
Removed WIP marker as OpenJ9 changes are now ready (eclipse/openj9#6098)

keithc-ca:
The invocation of `cmake` needs `-DOMR_DDR=$(OPENJ9_ENABLE_DDR)` else DDR will not be enabled when necessary leading to the error `No rule to make target 'j9ddr'` above.

Even with that, however, the build still fails to compile the generated DDR Java code.

dnakamura:
I beleieve the issue is that at the moment the cmake side is forcing `OMR_DDR=Off`  to work around the fact that OMR enables `OMR_DDR` by default.  It was a bit heavy handed to force the value off rather than default to off unless overridden on the command line. Updating the openj9 pr now

keithc-ca:
It isn't up to openj9 (nor should it be). A user should be able to choose `--enable-ddr` or `--disable-ddr` to suit their needs as they run `configure` in openjdk.

keithc-ca:
The default for OMR with autotools is to disable `OMR_DDR`; the default with cmake should be the same. OpenJ9 should respect the configuration choice from openjdk.

dnakamura:
@keithc-ca with this updated PR, and the update to the openj9 pr it should now be able to successfully build with ddr enabled.  As per our offline discussion:
- I am looking at changing the ddr implementation to no longer generate second project (thus harmonizing the build progress). However its looking to be a substantial refactor, and I don't think this PR should wait on that.

- At present the ddrgen is still a separate invocation of `make`. I thought that was the for the best at the moment since the build progress numbers are not harmonized which causes the %-age complete indication to bounce back and forth confusingly.

closed/OpenJ9.gmk
```diff
@@ -491,6 +496,12 @@ ifeq (true,$(OPENJ9_ENABLE_CMAKE))
     CMAKE_ARGS += -DCMAKE_C_COMPILER_LAUNCHER="$(ESCAPED_CCACHE)"
   endif # CCACHE
 
+  ifeq (true,$(OPENJ9_ENABLE_DDR))
+    CMAKE_ARGS += -DOMR_DDR=ON
+  else
+    CMAKE_ARGS += -DOMR_DDR=OFF
+  endif()
```
keithc-ca: `OPENJ9_ENABLE_DDR` is either `true` or `false` which I believe cmake understands, so this could be simplified to `-DOMR_DDR=$(OPENJ9_ENABLE_DDR)` in the initial definition of `CMAKE_ARGS`.

I'm not sure the parentheses are permitted on the `endif` line.

closed/OpenJ9.gmk
```diff
@@ -228,11 +228,16 @@ ifeq (true,$(OPENJ9_ENABLE_DDR))
...
+$(OPENJ9_VM_BUILD_DIR)/j9ddr.dat : run-ddrgen
+
 run-ddrgen :
+ifeq (true,$(OPENJ9_ENABLE_CMAKE))
+	$(MAKE) -C $(OPENJ9_VM_BUILD_DIR) j9ddr
+else
 	export CC="$(CC)" CXX="$(CXX)" $(EXPORT_MSVS_ENV_VARS) \
 		&& $(MAKE) -C $(OUTPUTDIR)/vm/ddr -f run_omrddrgen.mk
 endif
+endif
```
keithc-ca: Comments on these `endif` lines would be helpful to match them with their opening conditions.

dnakamura:
@keithc-ca  addressed your comments

closed/OpenJ9.gmk
```diff
@@ -228,11 +228,16 @@ ifeq (true,$(OPENJ9_ENABLE_DDR))
...
+
 run-ddrgen :
+ifeq (true,$(OPENJ9_ENABLE_CMAKE))
+	$(MAKE) -C $(OPENJ9_VM_BUILD_DIR) j9ddr
+else
 	export CC="$(CC)" CXX="$(CXX)" $(EXPORT_MSVS_ENV_VARS) \
 		&& $(MAKE) -C $(OUTPUTDIR)/vm/ddr -f run_omrddrgen.mk
-endif
+endif # ifeq (true,$(OPENJ9_ENABLE_CMAKE))
+endif # feq (true,$(OPENJ9_ENABLE_DDR))
```
keithc-ca: My preference is to just name the flag:
```
endif # OPENJ9_ENABLE_CMAKE
endif # OPENJ9_ENABLE_DDR
```
but if you want to repeat the condition, that's fine, just repeat all of it (`ifeq...` not `feq...`).

dnakamura:
@keithc-ca  fixed

keithc-ca:
Jenkins compile plinux,xlinuxcm jdk11

keithc-ca:
Apparently, PR builds are currently broken: [1]; built locally on xlinux with and without cmake.

[1] https://ci.eclipse.org/openj9/job/PullRequest-OpenJDK11/28/

PR #186, opened 2019-06-06
Propagate value of BOOT_JDK to cmake

keithc-ca:
CMake should use the tools in the boot jdk specified (or discovered) by `configure` rather than some other random JDK installed on the system; nor should it fail if a JDK cannot be found.


keithc-ca:
This must be merged before (or concurrently with) eclipse/openj9#6046.

PR #191, opened 2019-06-19
NativeDigest context free via Cleaner API

ashbm5:
due to issues such as eclipse/openj9#5611 removed the native context de-allocation via a finalizer. Introduced the use of PhantomReferences to control when the native digest context can be de-allocated.

The cleanup stage happens when the reference array is fully used, at which point there is an attempted clean-up by examining the ReferenceQueue which will de-allocate the native contexts and also free up space in the reference array.

Signed-off-by: Alon Shalev Housfater <alonsh@ca.ibm.com>

### DIFF
--- a/closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java
+++ b/closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java
@@ -38,6 +38,8 @@ import java.security.ProviderException;
 import static sun.security.provider.ByteArrayAccess.*;
 
 import jdk.crypto.jniprovider.NativeCrypto;
+import jdk.internal.ref.CleanerFactory;
+import java.lang.ref.Cleaner;
 
 abstract class NativeDigest extends MessageDigestSpi implements Cloneable {
 
@@ -55,10 +57,25 @@ abstract class NativeDigest extends MessageDigestSpi implements Cloneable {
     //  0: is already reset
     private long bytesProcessed;
 
-    private static NativeCrypto nativeCrypto;
+    private static final NativeCrypto nativeCrypto;
+    private static final Cleaner digestCleaner;
 
     static {
         nativeCrypto = NativeCrypto.getNativeCrypto();
+        digestCleaner = CleanerFactory.cleaner();
+    }
+
+    private static final class DigestCleanerRunnable implements Runnable {
+        private final long digestCtx;
+        
+        public DigestCleanerRunnable(long context) {
+            this.digestCtx = context;
+        }
+        
+        @Override
+        public void run() {
+            nativeCrypto.DigestDestroyContext(digestCtx);
+        }
     }
 
     /**
@@ -70,10 +87,12 @@ abstract class NativeDigest extends MessageDigestSpi implements Cloneable {
         this.digestLength = digestLength;
         this.algIndx = algIndx;
         this.context = nativeCrypto.DigestCreateContext(0, algIndx);
+
         if (this.context == -1) {
             throw new ProviderException("Error in Native Digest");
         }
 
+        digestCleaner.register(this, new DigestCleanerRunnable(this.context));
     }
 
     // return digest length. See JCA doc.
@@ -161,18 +180,13 @@ abstract class NativeDigest extends MessageDigestSpi implements Cloneable {
 
     synchronized public Object clone() throws CloneNotSupportedException {
         NativeDigest copy = (NativeDigest) super.clone();
-        copy.context    = nativeCrypto.DigestCreateContext(context, algIndx);
+        copy.context = nativeCrypto.DigestCreateContext(context, algIndx);
+
         if (copy.context == -1) {
             throw new ProviderException("Error in Native Digest");
         }
-        return copy;
-    }
 
-    /*
-     * Finalize method to release the Digest contexts.
-     */
-    @Override
-    public void finalize() {
-        nativeCrypto.DigestDestroyContext(context);
+        digestCleaner.register(copy, new DigestCleanerRunnable(copy.context));
+        return copy;
     }
 }
--- a/src/java.base/share/classes/sun/security/provider/SunEntries.java
+++ b/src/java.base/share/classes/sun/security/provider/SunEntries.java
@@ -90,12 +90,11 @@ final class SunEntries {
     /*
      * Check whether native crypto is enabled with property.
      * By default, the native crypto is enabled and uses native library crypto.
-     * The native crypto for MessageDigest is disabled until
-     * https://github.com/eclipse/openj9/issues/5611 can be resolved.
-     * The property 'jdk.nativeCrypto' is used to disable all native cryptos (Digest,
+     * The property 'jdk.nativeDigest' is used to disable Native digest alone
+     * and 'jdk.nativeCrypto' is used to disable all native cryptos (Digest,
      * CBC, GCM, RSA, and ChaCha20).
      */
-    private static boolean useNativeDigest = false;
+    private static boolean useNativeDigest = true;
 
     private SunEntries() {
         // empty


ashbm5:
@pshipton @DanHeidinga @keithc-ca still testing the changes locally. If possible, would like to get the review cycle started. See eclipse/openj9#5611 and eclipse/openj9#4530 for the actual error reports.

Also note that I can't be certain this PR will fix those issues as I'm unable to reproduce the error locally. However, looking at the code at Java_jdk_crypto_jniprovider_NativeCrypto_DigestUpdate which has extensive error checking likely the crash occurs here:
`(*OSSL_DigestUpdate)(context->ctx, (messageNative + messageOffset), messageLen)`

so I'm assuming context is de-allocated too early which is why the crash happens in the finalizer of `sun/security/ssl/BaseSSLSocketImpl`

pshipton:
This won't clean if new Digest's aren't being created. Ideally https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/ref/Cleaner.html should be used.

keithc-ca:
You may want to reuse the common cleaner accessible via `jdk.internal.ref.CleanerFactory.cleaner()`.

ashbm5:
@pshipton @keithc-ca is this also available for jdk8?

keithc-ca:
Yes, in Java 8 there is `sun.misc.Cleaner` (with differing API).

DanHeidinga:
@ashbm5 Are you adapting this PR in light of Keith & Peter's comments on using sun.misc.Cleaner?

ashbm5:
@DanHeidinga I'm a bit concerned about the performance impact of using sun.misc.Cleaner since it requires a background thread so I'm also evaluating the perf. impact of this approach. Also try different ways of resolving this issue - perhaps by buffering input data before sending to OpenSSL

DanHeidinga:
Look at how Cleaner is used in other places.  MutableCallSite is one example:
https://github.com/eclipse/openj9/blob/5d28821b3de3e0b01d0d9eaf2a8159628f819d0a/jcl/src/java.base/share/classes/java/lang/invoke/MutableCallSite.java#L84-L85

https://github.com/eclipse/openj9/blob/5d28821b3de3e0b01d0d9eaf2a8159628f819d0a/jcl/src/java.base/share/classes/java/lang/invoke/MutableCallSite.java#L294-L308

The Cleaner is processed immediately when it's being enqueued.  No extra thread is required.

ashbm5:
@pshipton @keithc-ca @DanHeidinga updated the PR with using the Cleaner API code. I did the relevant performance testing and found out that there is some impact when constant digest objects are being created and released when compared to finalizer but only slightly so while hopefully fixing the reported issues. with optimized code that doesn't constantly create new digest objects, there is no performance impact.

closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java
```diff
@@ -62,88 +57,24 @@
...
-          if (cryptoContextMgrSingleton == null) {
-              synchronized (NativeCryptoContextManager.class) {
-                if (cryptoContextMgrSingleton == null) {
-                    cryptoContextMgrSingleton = new NativeCryptoContextManager();
-                }
-              }
-          }
-          return cryptoContextMgrSingleton;
-    }
+    private static Cleaner digestCleaner;
```
keithc-ca: This field and `nativeCrypto` should both be `final`.

closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java
```diff
@@ -56,9 +57,24 @@
     private long bytesProcessed;
 
     private static NativeCrypto nativeCrypto;
+    private static Cleaner digestCleaner;
 
     static {
         nativeCrypto = NativeCrypto.getNativeCrypto();
+	digestCleaner = Cleaner.create();
```
keithc-ca: Do we *need* a distinct `Cleaner` object? Can we not reuse `CleanerFactory.cleaner()`?

ashbm5: updated

closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java
ashbm5: updated

closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java
```diff
@@ -154,9 +85,8 @@ private NativeCryptoContextReference(long ctx, NativeDigest referent, ReferenceQ
         this.algorithm = algorithm;
         this.digestLength = digestLength;
         this.algIndx = algIndx;
-
-        ctxMgr = getNativeCryptoContextManagerInstance();
-        this.context = ctxMgr.getContext(this,algIndx,0);
+        this.context = nativeCrypto.DigestCreateContext(0, algIndx);
+	digestCleaner.register(this,new DigestCleanerRunnable(this.context));
```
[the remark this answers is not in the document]
ashbm5: updated

src/java.base/share/classes/sun/security/provider/SunEntries.java
```diff
@@ -89,9 +89,8 @@
 
     /*
      * Check whether native crypto is enabled with property.
-     * By default, the native crypto is enabled  and uses native library crypto.
-     * The property 'jdk.nativeDigest' is used to disable Native digest alone
-     * and 'jdk.nativeCrypto' is used to disable all native cryptos (Digest,
+     * By default, the native crypto is enabled and uses native library crypto.
+     * The property 'jdk.nativeCrypto' is used to disable all native cryptos (Digest,
```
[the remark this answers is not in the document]
ashbm5: fixed

ashbm5:
@keithc-ca updated, please review

closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java
```diff
@@ -55,10 +57,25 @@
...
     static {
         nativeCrypto = NativeCrypto.getNativeCrypto();
+        digestCleaner = CleanerFactory.cleaner();
+    }
+
+    private static class DigestCleanerRunnable implements Runnable {
+        private long digestCtx;
+        
+        public DigestCleanerRunnable(long context) {
+               this.digestCtx = context;
```
keithc-ca: Please indent consistently (four spaces in this repository).

closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java
```diff
@@ -70,6 +87,8 @@
         this.digestLength = digestLength;
         this.algIndx = algIndx;
         this.context = nativeCrypto.DigestCreateContext(0, algIndx);
+        digestCleaner.register(this, new DigestCleanerRunnable(this.context));
```
keithc-ca: This should move after the error checking below.

closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java
```diff
@@ -162,17 +181,10 @@ synchronized protected final int engineDigest(byte[] out, int ofs, int len)
     synchronized public Object clone() throws CloneNotSupportedException {
         NativeDigest copy = (NativeDigest) super.clone();
         copy.context    = nativeCrypto.DigestCreateContext(context, algIndx);
+	digestCleaner.register(copy, new DigestCleanerRunnable(copy.context));
```
keithc-ca: This should move after the error checking below.

keithc-ca: @ashbm5 It appears you missed this comment.

ashbm5:
@keithc-ca updated, please review

src/java.base/share/classes/sun/security/provider/SunEntries.java
```diff
@@ -90,12 +90,10 @@
     /*
      * Check whether native crypto is enabled with property.
      * By default, the native crypto is enabled and uses native library crypto.
-     * The native crypto for MessageDigest is disabled until
-     * https://github.com/eclipse/openj9/issues/5611 can be resolved.
```
keithc-ca: Please restore the comment about `jdk.nativeDigest` from 31e38e7f9a.

closed/src/java.base/share/classes/sun/security/provider/NativeDigest.java
```diff
@@ -161,18 +180,13 @@ synchronized protected final int engineDigest(byte[] out, int ofs, int len)
 
     synchronized public Object clone() throws CloneNotSupportedException {
         NativeDigest copy = (NativeDigest) super.clone();
```
keithc-ca: I don't think using `super.clone()` is appropriate here: it will yield state that may be inconsistent with the newly allocated `context`. I suggest it should just be:
```java
    return new NativeDigest(algorithm, digestLength, algIndx);
```

ashbm5: that won't work, in the constructor `nativeCrypto.DigestCreateContext(0, algIndx)` is used while in clone() `nativeCrypto.DigestCreateContext(context, algIndx)` is needed to have openssl copy the current state of the cipher to the cloned cipher.

How about create an internal initCipher() method that will also take in the context and create an empty constructor. In the current constructor have initCipher() method called with context = 0 and in clone() invoke the empty constructor and then directly call initCipher() with the relevant context?

keithc-ca: Sorry, I missed the difference in the call to `DigestCreateContext` in `clone()` versus the constructor.
The only remaining change I request in `clone()` is to fix the indentation (no tabs in this repo).

src/java.base/share/classes/sun/security/provider/SunEntries.java
keithc-ca: @ashbm5 ?

keithc-ca:
Jenkins test all,sanity,extended zlinux jdk11

keithc-ca:
The build and both test jobs were successful, but something strange happened to the 'parallel' part of the test job: merging under the assumption that the strangeness is irrelevant.

PR #194, opened 2019-06-27
Put AArch64 j9jit29 back in java.base

knn-k:
This commit stops excluding j9jit29 in java.base when the target CPU is
aarch64.  It is to revert the temporary change in #126.

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -197,6 +197,7 @@ $(call openj9_add_jdk_shlibs, java.base, \
 	j9gc29 \
 	j9gcchk29 \
 	j9hookable29 \
+	j9jit29 \
 	j9jnichk29 \
 	j9jvmti29 \
 	j9prt29 \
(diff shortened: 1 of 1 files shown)


knn-k:
FYI @0xdaryl

0xdaryl:
@pshipton : could you review and merge please?

closed/OpenJ9.gmk
```diff
@@ -211,10 +212,6 @@ $(call openj9_add_jdk_shlibs, java.base, \
 	omrsig \
 	)
 
-ifneq (aarch64,$(OPENJDK_TARGET_CPU))
-  $(call openj9_add_jdk_shlibs, java.base, j9jit29)
```
0xdaryl: Isn't this line the one you want to do unconditionally now regardless of architecture?

pshipton: This line is replaced by adding j9jit29 to the list above.

PR #202, opened 2019-08-19
Change how Proxy ROMClasses are defined to store them in the SCC in Java 11

sogutbera:
Changes how Proxy ROMClasses are defined in Java 11 to how they are defined in Java 8 because they are being stored in the SCC in Java 8 but not in Java 11.

Closes: https://github.com/eclipse/openj9/issues/4985#

Signed-off-by: sogutbera <sogutbera@ibm.com>

### DIFF
--- a/closed/openjdk-tag.gmk
+++ b/closed/openjdk-tag.gmk
@@ -1,1 +1,1 @@
-OPENJDK_TAG := jdk-11.0.5+3
+OPENJDK_TAG := jdk-11.0.5+4


DanHeidinga:
This introduces a patch we'd need to carry forward for every future release.  Can you describe why changing to `defineClass0` is required?

sogutbera:
Because if Proxy ROMClasses are defined with Unsafe.defineClass, they will be tagged with an unsafe bit and isROMClassShareable will return false as mentioned in this comment: https://github.com/eclipse/openj9/issues/4985#issuecomment-470179477
But they are defined with defineClass0 in Java 8 and get stored in the SCC so that's why we can also use defineClass0 in Java 11.

sogutbera:
With this change, I am able to see the proxy class in the SCC as an orphan.

I am also able to see the traces mentioned in https://github.com/eclipse/openj9/issues/4985#: 
```
j9shr.1440     > API j9shr_classStoreTransaction_start : enter (classloader=0xbc0881f0 cpEntries=0x0 cpEntryCount=0 entryIndex=4294967295 loadType=1 partition=0x0 classname=$Proxy0 ismodified=0 takeReadWriteLock=1)
...
j9shr.1521     - CM commitOrphanROMClass : Data was stored in the cache for J9ROMClass $Proxy0 at address 0x9d6705b8
...
j9shr.1465     - API j9shr_classStoreTransaction_stop : New data has been stored to the cache for class $Proxy0 (ROMClass address 0xa5670590).
```
I am also seeing AOT warm's in the first run: 
```
(AOT warm) $Proxy0.originalMethod(Ljava/lang/String;)V @ 00007F6DBA2D7860-00007F6DBA2D7C15 OrdinaryMethod - Q_SZ=0 Q_SZI=0 QW=6 j9m=0000000001D6B338 bcsz=34 sync GCR compThread=0 CpuLoad=114%(28%avg) JvmCpu=98%Stored AOT code for ROMMethod 0x00007F6D816709A0 in shared cache.
```
and AOT load's in the subsequent runs:
```
(AOT load) $Proxy0.originalMethod(Ljava/lang/String;)V @ 00007FCC3E896B20-00007FCC3E896ED5 Q_SZ=0 Q_SZI=0 QW=1 j9m=0000000002451338 bcsz=34 compThread=0Found AOT code for ROMMethod 0x00007FCC0565CDC4 in shared cache.
```

DanHeidinga:
@sogutbera I'm glad to see this works and shows that the AOT code can still be loaded for Proxy classes.  Unfortunately it adds a patch to OpenJDK code that we'd have a hard time getting upstreamed and would make our implementation and RI diverge here.

A better approach is to figure out if we can allow classes defined by `Unsafe.defineClass` or `Unsafe.defineAnonymousClass` to be saved in the SCC as we're likely to see code generators in the JDK standardize on those APIs.  Better to solve the problem once now then to keep punting it down the road.

hangshao0:
We may consider adding an new command line option to allow unsafe classes to be stored to the shared cache.

PR #205, opened 2019-09-06
Compile DDR_VM code twice

keithc-ca:
* once with generated source files
* again with generated class files to validate blob integrity
* stop staging debugtools
* merge DDR support makefiles

Once all Java versions validate their blobs this way, a change can be introduced in openj9 that distinguishes core files created after a certain time allowing pointer classes to be derived at runtime, fixing eclipse/openj9#2507.

See also ibmruntimes/openj9-openjdk-jdk8#117.


pshipton:
jenkins test sanity win jdk11

PR #207, opened 2019-09-11
Correct path to j9ddr.dat in cmake builds

keithc-ca:


keithc-ca:
This should fix travis failures, e.g. https://travis-ci.org/eclipse/openj9/jobs/583845080

PR #209, opened 2019-09-17
Modify RSA Key for multithread performance

shanchao95:
*add queue for rsa key
*remove synchronize for rsa operation

Signed-off-by: Chao Shan <chao.shan@ibm.com>

### DIFF
--- a/closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
+++ b/closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
@@ -94,23 +94,28 @@ public final class NativeRSACore {
     /**
      * RSA public key ops. Simple modPow().
      */
-    synchronized private static byte[] crypt_Native(byte[] msg, sun.security.rsa.RSAPublicKeyImpl key)
+    private static byte[] crypt_Native(byte[] msg, sun.security.rsa.RSAPublicKeyImpl key)
         throws BadPaddingException {
-
         long nativePtr = key.getNativePtr();
-
         if (nativePtr == -1) {
             return null;
         }
 
-        BigInteger n = key.getModulus();
-        byte[] output = new byte[getByteLength(n)];
+        byte[] output;
+        int outputLen;
 
-        int outputLen = nativeCrypto.RSAEP(msg, msg.length, output, nativePtr);
+        try {
+            BigInteger n = key.getModulus();
+            output = new byte[getByteLength(n)];
+            outputLen = nativeCrypto.RSAEP(msg, msg.length, output, nativePtr);
+        } finally {
+            key.returnNativePtr(nativePtr);
+        }
 
         if (outputLen == -1) {
             return null;
         }
+
         return output;
     }
 
@@ -118,26 +123,25 @@ public final class NativeRSACore {
      * RSA private key operations with CRT. Algorithm and variable naming
      * are taken from PKCS#1 v2.1, section 5.1.2.
      */
-    synchronized private static byte[] crtCrypt_Native(byte[] msg, sun.security.rsa.RSAPrivateCrtKeyImpl key,
+    private static byte[] crtCrypt_Native(byte[] msg, sun.security.rsa.RSAPrivateCrtKeyImpl key,
             boolean verify) throws BadPaddingException {
         long nativePtr = key.getNativePtr();
-
         if (nativePtr == -1) {
             return null;
         }
 
-        int verifyInt;
-        BigInteger n = key.getModulus();
-        int outputLen = getByteLength(n);
-        byte[] output = new byte[outputLen];
+        int outputLen;
+        byte[] output;
 
-        if(verify) {
-            verifyInt = outputLen;
-        } else {
-            verifyInt = -1;
+        try {
+            BigInteger n = key.getModulus();
+            outputLen = getByteLength(n);
+            output = new byte[outputLen];
+            int verifyInt = verify ? outputLen : -1;
+            outputLen = nativeCrypto.RSADP(msg, msg.length, output, verifyInt, nativePtr);
+        } finally {
+            key.returnNativePtr(nativePtr);
         }
-
-        outputLen = nativeCrypto.RSADP(msg, msg.length, output, verifyInt, nativePtr);
 
         if (outputLen == -1) {
             return null;
--- a/src/java.base/share/classes/sun/security/rsa/RSAPrivateCrtKeyImpl.java
+++ b/src/java.base/share/classes/sun/security/rsa/RSAPrivateCrtKeyImpl.java
@@ -32,6 +32,7 @@ package sun.security.rsa;
 
 import java.io.IOException;
 import java.math.BigInteger;
+import java.util.concurrent.ConcurrentLinkedQueue;
 
 import java.security.*;
 import java.security.spec.*;
@@ -62,6 +63,8 @@ public final class RSAPrivateCrtKeyImpl
         extends PKCS8Key implements RSAPrivateCrtKey {
 
     private static final long serialVersionUID = -1326088454257084918L;
+
+    private final ConcurrentLinkedQueue<Long> keyQ = new ConcurrentLinkedQueue<>();
 
     private BigInteger n;       // modulus
     private BigInteger e;       // public exponent
@@ -245,16 +248,7 @@ public final class RSAPrivateCrtKeyImpl
         return coeff;
     }
 
-    private long nativeRSAKey = 0x0;
-
-    /**
-     * Get native RSA Public Key context pointer. 
-     * Create native context if uninitialized.
-     */
-    protected long getNativePtr() {
-        if (nativeRSAKey != 0x0) {
-            return nativeRSAKey;
-        }
+    private long RSAPrivateKey_generate() {
 
         BigInteger n =    this.getModulus();
         BigInteger d =    this.getPrivateExponent();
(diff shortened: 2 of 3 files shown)


ashbm5:
@keithc-ca 
This is a performance enhancement for RSA Sign/Verify/Decrypt/Encrypt operations. Running a multi-threaded RSA workload indicates a synchronized statement prevents any multi-threaded operations of the RSA code. The synchronized was removed and a queue mechanism was added to prevent any multi-threading issues when using the OpenSSL code. Specifically, the queue was added so that a single RSA key cannot be concurrently used by two concurrent RSA operations.

The problematic situation is as follows:
if there are two java.security.Signature objects where the following occurs:
1) initialized via initSign() using the same PrivateKey object in two different threads and then sign() called concurrently.
or
2) initialized via initVerify() using the same PublicKey object in two different threads and then verify() called concurrently.

It is possible that the corresponding OpenSSL routines for sign/verify will be called concurrently on the same OpenSSL native key. While we didn't attempt to produce a failure - OpenSSL didn't clearly document the multi-threaded safety of these routines. see ``https://www.openssl.org/blog/blog/2017/02/21/threads/`` and also ``https://openssl-users.openssl.narkive.com/UW2ozVNB/thread-safety-of-crypto-library`` which indicates there is (or was) a multi-threading issue with RSA.

See the blog which states `` Almost none of this behavior is documented, and it’s risky to plan on future behavior from just looking at the current code``. Also, it is indicated that different openssl versions may handle threads differently and OpenJ9 supports OpenSSL 1.0.2 to OpenSSL 1.1.1. Therefore, the code was added for safety and compatibility.

src/java.base/share/classes/sun/security/rsa/RSAPrivateCrtKeyImpl.java
```diff
@@ -62,6 +63,8 @@
         extends PKCS8Key implements RSAPrivateCrtKey {
 
     private static final long serialVersionUID = -1326088454257084918L;
+    
+    private final ConcurrentLinkedQueue<Long> keyQ = new ConcurrentLinkedQueue<Long>();
```
keithc-ca: nit: don't repeat type argument

src/java.base/share/classes/sun/security/rsa/RSAPrivateCrtKeyImpl.java
```diff
@@ -276,16 +271,32 @@ protected long getNativePtr() {
...
+        if (ptr != null) {
+            keyQ.add(ptr);
+        } 
     }
 
     @Override
     public void finalize() {
-        if (nativeRSAKey != 0x0 && nativeRSAKey != -1) {
-            nativeCrypto.destroyRSAKey(nativeRSAKey);
+        Long itr = null;
```
keithc-ca: Please remove redundant initialization (javac will ensure it's assigned a value before use).

src/java.base/share/classes/sun/security/rsa/RSAPublicKeyImpl.java
```diff
@@ -60,6 +61,8 @@
 
     private static final long serialVersionUID = 2644735423591199609L;
     private static final BigInteger THREE = BigInteger.valueOf(3);
+    
+    private final ConcurrentLinkedQueue<Long> keyQ = new ConcurrentLinkedQueue<Long>();
```
keithc-ca: Please remove redundant type argument.

src/java.base/share/classes/sun/security/rsa/RSAPublicKeyImpl.java
```diff
@@ -217,31 +220,37 @@ protected Object writeReplace() throws java.io.ObjectStreamException {
...
+    byte[] e_2c = e.toByteArray();
 
-        BigInteger n = this.getModulus();
-        BigInteger e = this.getPublicExponent();
-
-        byte[] n_2c = n.toByteArray();
-        byte[] e_2c = e.toByteArray();
+    newNativePtr = nativeCrypto.createRSAPublicKey(n_2c,n_2c.length, e_2c, e_2c.length);
+    return newNativePtr;
+}
```
keithc-ca: Please restore proper indentation.

src/java.base/share/classes/sun/security/rsa/RSAPublicKeyImpl.java
```diff
@@ -217,31 +220,37 @@ protected Object writeReplace() throws java.io.ObjectStreamException {
...
+    if (ptr != null) {
+        keyQ.add(ptr);
+    } 
+}
+
+@Override
     public void finalize() {
-        if (nativeRSAKey != 0x0 && nativeRSAKey != -1) {
-           nativeCrypto.destroyRSAKey(nativeRSAKey);
+        Long itr = null;
```
keithc-ca: Please remove redundant initialization.

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
```diff
@@ -94,51 +94,63 @@ public static int getByteLength(BigInteger b) {
...
+        throws BadPaddingException {  
         long nativePtr = key.getNativePtr();
-
         if (nativePtr == -1) {
             return null;
         }
 
-        BigInteger n = key.getModulus();
-        byte[] output = new byte[getByteLength(n)];
+        BigInteger n;
```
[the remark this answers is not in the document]
shanchao95: Hello Sir, those variables are used outside of `try` block.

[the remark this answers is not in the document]
keithc-ca: I said 'this' because I was only talking about one variable (`n`) which is only used _within_ the `try` block.

src/java.base/share/classes/sun/security/rsa/RSAPublicKeyImpl.java
```diff
@@ -217,31 +220,37 @@ protected Object writeReplace() throws java.io.ObjectStreamException {
...
-    /**
-     * Get native RSA Public Key context pointer.
-     * Create native context if uninitialized.
-     */
-    protected long getNativePtr() {
-        if (nativeRSAKey != 0x0) {
-            return nativeRSAKey;
-        }
-
+    private long RSAPrivateKey_generate() {
```
keithc-ca: The name of this method is misleading: it generates the public, key - not the private key.

src/java.base/share/classes/sun/security/rsa/RSAPrivateCrtKeyImpl.java
```diff
@@ -276,16 +271,32 @@ protected long getNativePtr() {
...
+
+    protected long getNativePtr() {
+        Long ptr = keyQ.poll();
+        if (ptr == null) {
+            return RSAPrivateKey_generate();
+        }
+        return ptr;
+    }
+
+    protected void returnNativePtr(Long ptr) {
```
keithc-ca: The way this is used, `ptr` will never be `null`. I suggest you change the parameter type to `long` (to match the caller) and do the boxing here. Same comment applies to similar code in `RSAPublicKeyImpl`.

shanchao95: Hello Sir, sorry i am not sure if i understand boxing correctly. is this what you suggest for the modification? 
`protected void returnNativePtr(long ptr) {`
      `      keyQ.add(ptr);`
`    }`

keithc-ca: Yes, that's what I meant. There's an implicit use of `Long.valueOf()` in `keyQ.add(ptr);` so it's as if you wrote `keyQ.add(Long.valueOf(ptr));`.

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
```diff
@@ -94,51 +94,61 @@ public static int getByteLength(BigInteger b) {
...
-            verifyInt = -1;
+        int outputLen;
+        byte[] output;
+
+        try {
+            int verifyInt;
+            BigInteger n = key.getModulus();
+            outputLen = getByteLength(n);
+            output = new byte[outputLen];
+            if (verify) {
```
keithc-ca: This might be easier to read if `verifyInt` were declared and initialized here as:
```
int verifyInt = verify ? outputLen : -1;
```

src/java.base/share/classes/sun/security/rsa/RSAPrivateCrtKeyImpl.java
```diff
@@ -276,16 +270,28 @@ protected long getNativePtr() {
...
+    protected long getNativePtr() {
+        Long ptr = keyQ.poll();
+        if (ptr == null) {
+            return RSAPrivateKey_generate();
+        }
+        return ptr;
+    }
+
+    protected void returnNativePtr(long ptr) {
+            keyQ.add(ptr);
```
keithc-ca: Please fix the indentation (just 4 spaces, not 8).

src/java.base/share/classes/sun/security/rsa/RSAPrivateCrtKeyImpl.java
```diff
@@ -276,16 +270,28 @@ protected long getNativePtr() {
         byte[] dQ_2c   = dQ.toByteArray();
         byte[] qInv_2c = qInv.toByteArray();
 
-        nativeRSAKey = nativeCrypto.createRSAPrivateCrtKey(n_2c,n_2c.length, d_2c, d_2c.length, e_2c, e_2c.length,
+        return nativeCrypto.createRSAPrivateCrtKey(n_2c,n_2c.length, d_2c, d_2c.length, e_2c, e_2c.length,
```
keithc-ca: Formatting: space after comma.

src/java.base/share/classes/sun/security/rsa/RSAPublicKeyImpl.java
```diff
@@ -217,31 +220,33 @@ protected Object writeReplace() throws java.io.ObjectStreamException {
...
+    private long RSAPublicKey_generate() {
         BigInteger n = this.getModulus();
         BigInteger e = this.getPublicExponent();
 
         byte[] n_2c = n.toByteArray();
         byte[] e_2c = e.toByteArray();
 
-        nativeRSAKey = nativeCrypto.createRSAPublicKey(n_2c,n_2c.length, e_2c, e_2c.length);
-        return nativeRSAKey;
+        return nativeCrypto.createRSAPublicKey(n_2c,n_2c.length, e_2c, e_2c.length);
```
keithc-ca: Formatting: space after comma.

src/java.base/share/classes/sun/security/rsa/RSAPublicKeyImpl.java
```diff
@@ -217,31 +220,33 @@ protected Object writeReplace() throws java.io.ObjectStreamException {
...
+    protected long getNativePtr() {
+        Long ptr = keyQ.poll();
+        if (ptr == null) {
+            return RSAPublicKey_generate();
+        }
+        return ptr;
+    }
+
+    protected void returnNativePtr(long ptr) {
+            keyQ.add(ptr);
```
keithc-ca: Please fix indentation.

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
```diff
@@ -94,51 +94,55 @@ public static int getByteLength(BigInteger b) {
...
-
-        if(verify) {
-            verifyInt = outputLen;
-        } else {
-            verifyInt = -1;
+        int outputLen;
+        byte[] output;
+
+        try {
+            int verifyInt = verify ? outputLen : -1;
```
keithc-ca: This won't compile: `outputLen` doesn't have a value yet.

keithc-ca:
Before I merge this, please open pull requests for openj9-openjdk-jdk8, openj9-openjdk-jdk13 and openj9-openjdk-jdk so all supported versions of Java get the same improvement.

PR #211, opened 2019-10-01
Include libbcvrelationships.so for class relationships tests

sharon-wang:
For PR: https://github.com/eclipse/openj9/pull/7306


keithc-ca:
This can be merged safely in anticipation of eclipse/openj9#7306 - the file will only be copied if it exists.

PR #212, opened 2019-10-02
Fix openj9 javadoc creation

lumpfish:
Signed-off-by: Simon Rushton <rushton@uk.ibm.com>

### DIFF
--- a/closed/make/Docs.gmk
+++ b/closed/make/Docs.gmk
@@ -1,5 +1,5 @@
 # ===========================================================================
-# (c) Copyright IBM Corp. 2018, 2018 All Rights Reserved
+# (c) Copyright IBM Corp. 2018, 2019 All Rights Reserved
 # ===========================================================================
 # 
 # This code is free software; you can redistribute it and/or modify it
@@ -21,3 +21,100 @@
 # ===========================================================================
 
 JVMTI_HTML := $(TOPDIR)/closed/variant-server/gensrc/jvmtifiles/jvmti.html
+
+##############################################################################
+# Text for openj9 modules
+
+OPENJ9_BASE_URL := https://www.eclipse.org/openj9/
+OPENJ9_BUG_SUBMIT_URL := https://github.com/eclipse/openj9/issues
+OPENJ9_FULL_COMPANY_NAME := IBM Corp
+OPENJ9_JAVADOC_BOTTOM := \
+    <a href="$(OPENJ9_BASE_URL)">Link</a> to Eclipse OpenJ9 documentation.<br> \
+    To raise a bug report or suggest an improvement create an <a href="$(OPENJ9_BUG_SUBMIT_URL)">Eclipse Openj9 issue.</a><br> \
+    Copyright &copy; 1998, $(COPYRIGHT_YEAR), $(OPENJ9_FULL_COMPANY_NAME). All rights reserved.
+
+##############################################################################
+# The javadoc tool only allows for a single setting for headers and footers
+# per invocation. The sed command below replaces the footer for the pages
+# relating to openj9 classes to the correct value for those classes.
+#
+# '.' need escaping in the sed regex. If more special characters are introduced
+# in the openjdk JAVADOC_BOTTOM in future these would need escaping also.
+#
+# '&' needs escaping in sed replacement text.
+#
+# {@docroot} is expanded by javadoc when the html files are created, so the matching
+# needs to cater for anything {@docroot} may have been substituted by - '../../../..'
+# or '../..' or '..' etc.
+
+# \(\.\./\)*\.\. means zero  or more '../' strings followed by '..'
+
+JAVADOC_DOCROOT := \(\.\./\)*\.\.
+JAVADOC_BOTTOM_REGEX = $(subst {@docroot},$(JAVADOC_DOCROOT),$(subst .,\.,$(JAVADOC_BOTTOM)))
+OPENJ9_JAVADOC_BOTTOM := $(subst &,\&,$(OPENJ9_JAVADOC_BOTTOM))
+OPENJ9_SUBST_COMMAND = $(SED) -e '\''s|$(JAVADOC_BOTTOM_REGEX)|$(OPENJ9_JAVADOC_BOTTOM)|g'\''
+
+################################################################################
+# Define a javadoc Group for the OpenJ9 specific modules
+# Module names are added to DOCS_MODULES in closed/make/common/Modules.gmk
+OPENJ9_GROUP_NAME := OpenJ9
+OPENJ9_GROUP_MODULES := openj9.*
+OPENJ9_GROUP_DESCRIPTION := \
+    Modules whose names start with {@code openj9.} contain APIs provided by the \
+    <a href="$(OPENJ9_JAVADOC_BASE_URL)">Eclipse OpenJ9</a> project. \
+    These APIs can only be used with a JDK which includes the OpenJ9 virtual machine. \
+    #
+JDK_GROUPS += OPENJ9
+
+################################################################################
+# Add openj9 specific modules to the list of modules for which javadoc is required
+OPENJ9_MODULES := $(filter openj9.%,$(call FindAllModules))
+DOCS_MODULES += $(OPENJ9_MODULES)
+
+OPENJDK_JAVADOC_MARKER := $(DOCS_OUTPUTDIR)/api/index.html
+OPENJ9_JAVADOC_MARKER := $(SUPPORT_OUTPUTDIR)/docs/_javadoc_JDK_API_openj9.marker
+JDK_API_CUSTOM_TARGETS += $(OPENJ9_JAVADOC_MARKER)
+
+OPENJ9_JAVADOC_SCRIPT := $(SUPPORT_OUTPUTDIR)/docs/_javadoc_openj9_footer_replace.sh
+OPENJ9_JAVADOC_TEMP  := $(SUPPORT_OUTPUTDIR)/docs/_javadoc_temp.html
+
+################################################################################
+# The OpenJ9 classes are built from the j9jcl_sources and closed/src
+# directories.
+# The packages are beneath directories called /share/classes (or /unix/classes,
+# /windows/classes etc.) which do not appear in the javadoc tree structure.
+# The commands below replace the footer in the javadoc html pages for the
+# following files:
+# 1. The openj9 java source files which override the openjdk versions
+# 2. The openj9.xxxx modules, including the javadoc generated files such as
+#    enum class pages, module-summary.html, package-summary.html etc.)
+# 3. The com/ibm/xxxx packages, including the javadoc generated files such as
+#    enum class pages, module-summary.html, package-summary.html etc.)
+# A temporary script file is used to avoid xargs command line length issues
+# on some platforms. 
+
+OPENJ9_JAVADOC_SCRIPT_BODY = \
+    for path in "$$@" ; do \
+      if [ -f "$(DOCS_OUTPUTDIR)/api/$$path" ] ; then \
+        $(CP) "$(DOCS_OUTPUTDIR)/api/$$path" "$(OPENJ9_JAVADOC_TEMP)" ; \
+        $(OPENJ9_SUBST_COMMAND) < "$(OPENJ9_JAVADOC_TEMP)" > "$(DOCS_OUTPUTDIR)/api/$$path" ; \
+        $(RM) "$(OPENJ9_JAVADOC_TEMP)" ; \
+      fi ; \
+    done \
+    #
+
+$(OPENJ9_JAVADOC_MARKER) : $(OPENJDK_JAVADOC_MARKER)
+	$(ECHO) '$(OPENJ9_JAVADOC_SCRIPT_BODY)' >$(OPENJ9_JAVADOC_SCRIPT)
+	$(CD) $(SUPPORT_OUTPUTDIR)/j9jcl_sources ; \
+		$(FIND) . -type f -name '*.java' \
+			| $(SED) -e 's|/[^/]\+/classes/|/|' -e 's|\.java$$|.html|' \
+			| $(XARGS) $(BASH) $(OPENJ9_JAVADOC_SCRIPT)
+	$(CD) $(TOPDIR)/closed/src ; \
+		$(FIND) . -type f -name '*.java' \
+			| $(SED) -e 's|/[^/]\+/classes/|/|' -e 's|\.java$$|.html|' \
+			| $(XARGS) $(BASH) $(OPENJ9_JAVADOC_SCRIPT)
+	$(CD) $(DOCS_OUTPUTDIR)/api ; \
+		$(FIND) . -type f '(' -path '*/openj9*/*.html' ')' -o '(' -path '*/com/ibm/*.html' ')' \
+			| $(XARGS) $(BASH) $(OPENJ9_JAVADOC_SCRIPT)
+	$(RM) -f $(OPENJ9_JAVADOC_SCRIPT)
+	$(TOUCH) $@
--- a/closed/make/common/Modules.gmk
+++ b/closed/make/common/Modules.gmk
@@ -1,5 +1,5 @@
 # ===========================================================================
-# (c) Copyright IBM Corp. 2017, 2018 All Rights Reserved
+# (c) Copyright IBM Corp. 2017, 2019 All Rights Reserved
 # ===========================================================================
 # This code is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License version 2 only, as
@@ -29,6 +29,9 @@ MODULES_FILTER += \
     jdk.internal.vm.ci \
     jdk.internal.vm.compiler \
     jdk.internal.vm.compiler.management \
+    jdk.jfr \
+    jdk.jstatd \
+    jdk.management.jfr \
     #
 
 TOP_SRC_DIRS += \


closed/make/Docs.gmk
```diff
@@ -1,5 +1,5 @@
 # ===========================================================================
-# (c) Copyright IBM Corp. 2018, 2018 All Rights Reserved
+# (c) Copyright IBM Corp. 2018, 2019 All Rights Reserved
```
pshipton: Is this file derived from OpenJDK content?

closed/make/Docs.gmk
```diff
@@ -20,4 +20,115 @@
...
+  CMD_OUTPUT = $(shell if [ -f $(OPENJDK_SUN_MGMT_SRCDIR)/GarbageCollectorMXBean.java ]; then $(MV) $(OPENJDK_SUN_MGMT_SRCDIR)/GarbageCollectorMXBean.java $(OPENJDK_SUN_MGMT_SRCDIR)/GarbageCollectorMXBean.javaz; fi)
+  CMD_OUTPUT = $(shell if [ -f $(OPENJDK_SUN_MGMT_SRCDIR)/OperatingSystemMXBean.java ]; then $(MV) $(OPENJDK_SUN_MGMT_SRCDIR)/OperatingSystemMXBean.java $(OPENJDK_SUN_MGMT_SRCDIR)/OperatingSystemMXBean.javaz; fi)
+  CMD_OUTPUT = $(shell if [ -f $(OPENJDK_SUN_MGMT_SRCDIR)/UnixOperatingSystemMXBean.java ]; then $(MV) $(OPENJDK_SUN_MGMT_SRCDIR)/UnixOperatingSystemMXBean.java $(OPENJDK_SUN_MGMT_SRCDIR)/UnixOperatingSystemMXBean.javaz; fi)
+  CMD_OUTPUT = $(shell if [ -f $(OPENJDK_SUN_MGMT_SRCDIR)/package-info.java ]; then $(MV) $(OPENJDK_SUN_MGMT_SRCDIR)/package-info.java $(OPENJDK_SUN_MGMT_SRCDIR)/package-info.javaz; fi)
+  # Then rename any remaining files we don't want to generate javadoc for
+  # This finds all files ending in .java and then renames each one with an 'x' added add its last character.
+  CMD_OUTPUT = $(shell find $(OPENJDK_SUN_MGMT_SRCDIR) -maxdepth 1 -type 'f' -name "*.java" | xargs -I {} sh -c 'set -x; $(MV) $$1 $${1}x; set +x' -- {})
```
keithc-ca: Please remove unnecessary quotes (e.g. `'f'`, `"/openj9"`, `".html"`).

closed/make/Docs.gmk
```diff
@@ -20,4 +20,115 @@
...
+  CMD_OUTPUT = $(shell if [ -f $(OPENJ9_SUN_MGMT_SRCDIR)/HotSpotDiagnosticMXBean.java ]; then $(MV) $(OPENJ9_SUN_MGMT_SRCDIR)/HotSpotDiagnosticMXBean.java $(OPENJ9_SUN_MGMT_SRCDIR)/HotSpotDiagnosticMXBean.javax; fi)
+  # Finally restore the .java extension of the files we do want to generate javadoc for and the package-info.java file if present
+  CMD_OUTPUT = $(shell find $(OPENJDK_SUN_MGMT_SRCDIR) -maxdepth 1 -type 'f' -name "*.javaz" | xargs -I {} sh -c 'set -x; $(MV) $$1 $${1%?}; set +x' -- {})
+endif
+
+# Target to change the web page footers to the required values.
+# This ensures that openj9 and openjdk classes contain the correct copyrights etc.
+openj9-docs-set-javadoc-bottom:
+	# Set footer for openj9 modules, packages and classes
+	cd $(DOCS_OUTPUTDIR)/api; find . -type 'f' | grep "/openj9" | grep ".html" | xargs -I {} $(OPENJ9_SUBST_COMMAND) {}
```
keithc-ca: `grep /openj9 | grep .html` matches too many paths;
I think you want something more like `grep '/openj9\\..*\\.html$'.

closed/make/Docs.gmk
```diff
@@ -20,4 +20,115 @@
...
+	cd $(DOCS_OUTPUTDIR)/api; find . -type 'f' | xargs -I {} $(OPENJDK_SUBST_COMMAND) {}
+
+openj9-docs-restore-files:
+	# This finds all files ending in .javax and then renames each one with its last character removed.
+	find $(OPENJDK_SUN_MGMT_SRCDIR) -maxdepth 1 -type 'f' -name "*.javax" | xargs -I {} sh -c 'set -x; $(MV) $$1 $${1%?}; set +x' -- {}
+	find $(OPENJ9_SUN_MGMT_SRCDIR) -maxdepth 1 -type 'f' -name "*.javax" | xargs -I {} sh -c 'set -x; $(MV) $$1 $${1%?}; set +x' -- {}
+	# Also restore any files ending in .javaz which would exist if a previous run was interrupted in the small window when those files were renamed
+	find $(OPENJDK_SUN_MGMT_SRCDIR) -maxdepth 1 -type 'f' -name "*.javaz" | xargs -I {} sh -c 'set -x; $(MV) $$1 $${1%?}; set +x' -- {}
+
+PHONY: openj9-docs-set-javadoc-bottom openj9-docs-restore-files
```
keithc-ca: This needs to be `.PHONY` (the `.` is missing).

make/Docs.gmk
```diff
@@ -37,7 +41,7 @@ $(eval $(call ReadImportMetaData))
 
 ################################################################################
 # Hook to include the corresponding custom file, if present.
-$(eval $(call IncludeCustomExtension, Docs.gmk))
+#$(eval $(call IncludeCustomExtension, Docs.gmk))
```
keithc-ca: It's disappointing that this file needs to change. Did you try alternative approaches in `close/make/Docs.gmk`?

closed/make/Docs.gmk
```diff
@@ -20,4 +20,115 @@
...
+OPENJDK_SUN_MGMT_SRCDIR = $(OPENJDK_JCL_TOPDIR)/jdk.management/share/classes/jdk.management/com/sun/management
+OPENJ9_SUN_MGMT_SRCDIR = $(OPENJ9_JCL_TOPDIR)/jdk.management/share/classes/com/sun/management
+
+JDK_API_CUSTOM_TARGETS += openj9-docs-set-javadoc-bottom openj9-docs-restore-files
+
+# Code should only be executed when generating the docs-jdk-api-javadoc target.
+# If executed for any other targets the renames would be done but the target
+# to restore the files would not be executed.
+
+ifneq (,$(findstring docs-jdk-api-javadoc,$(MAKECMDGOALS)))
```
keithc-ca: What about other targets of `Docs.gmk`, like these:
```
docs-javase-api-javadoc
docs-javase-api-modulegraph
docs-jdk-api-modulegraph
docs-jdk-index
docs-jdk-specs
docs-reference-api-javadoc
docs-reference-api-modulegraph
docs-zip
```

keithc-ca:
This doesn't work for me; `make && make docs-jdk-api-javadoc` yields
```
Building target 'docs-jdk-api-javadoc' in configuration 'linux-x86_64-normal-server-release'
[...snip...]
Generating JDK_API javadoc for 65 modules
/home/keithc/space/jdk11/closed/make/Docs.gmk:117: recipe for target 'openj9-docs-set-javadoc-bottom' failed
make/Main.gmk:364: recipe for target 'docs-jdk-api-javadoc' failed
make[2]: *** [docs-jdk-api-javadoc] Error 1

ERROR: Build failed for target 'docs-jdk-api-javadoc' in configuration 'linux-x86_64-normal-server-release' (exit code 2) 

=== Make failed targets repeated here ===
/home/keithc/space/jdk11/closed/make/Docs.gmk:117: recipe for target 'openj9-docs-set-javadoc-bottom' failed
make/Main.gmk:364: recipe for target 'docs-jdk-api-javadoc' failed
=== End of repeated output ===
```

closed/make/Docs.gmk
lumpfish: No.  The only knowledge derived from OpenJDK are are the names of the macros whose values are overridden.

closed/make/Docs.gmk
```diff
@@ -20,4 +20,115 @@
...
+# These placeholders are replaced with the correct (Oracle, IBM, etc.)
+# copyrights after the javadoc has been generated.
+# The javadoc tool only allows for a single setting for headers and footers
+# per invocation, so this hack allows for all the javadoc pages to be linked
+# from a single top level index.html page with the copyrights set correctly
+# for individual API pages.
+
+OPENJDK_JAVADOC_BOTTOM := $(JAVADOC_BOTTOM)
+OPENJDK_SUBST_COMMAND := sed -i 's|JAVADOC_BOTTOM|$(OPENJDK_JAVADOC_BOTTOM)|g'
+OPENJDK_JCL_TOPDIR=$(SUPPORT_OUTPUTDIR)/src/src
```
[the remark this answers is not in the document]
lumpfish: Will change the erroneous '=' to be ':='.

closed/make/Docs.gmk
```diff
@@ -20,4 +20,115 @@
...
+OPENJ9_SUBST_COMMAND := $(subst &,\&,$(OPENJ9_SUBST_COMMAND))
+OPENJDK_SUBST_COMMAND := $(subst &,\&,$(OPENJDK_SUBST_COMMAND))
+
+################################################################################
+# Define a javadoc Group for the OpenJ9 specific modules
+# Module names are added to DOCS_MODULES in closed/make/common/Modules.gmk
+OpenJ9_GROUP_NAME := OpenJ9
+OpenJ9_GROUP_MODULES := openj9.*
+OpenJ9_GROUP_DESCRIPTION := \
+    Modules whose names start with {@code openj9}. contain APIs provided by the \
```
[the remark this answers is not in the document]
lumpfish: Will fix to move the period inside the `{@code}` element.

closed/make/Docs.gmk
```diff
@@ -20,4 +20,115 @@
...
+#    used) with an openj9 based jdk.
+# The code below renames the files which cannot be used while javadoc is
+# executed so no javadoc is generated for those classes.
+#
+# This is not done within a target because the javadoc dependencies are generated
+# from the source tree before executing any targets, so dependencies would be generated
+# for files found before they are renamed, causing build errors.  
+
+OPENJDK_SUN_MGMT_SRCDIR = $(OPENJDK_JCL_TOPDIR)/jdk.management/share/classes/jdk.management/com/sun/management
+OPENJ9_SUN_MGMT_SRCDIR = $(OPENJ9_JCL_TOPDIR)/jdk.management/share/classes/com/sun/management
```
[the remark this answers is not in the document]
lumpfish: Changed to :=

closed/make/Docs.gmk
lumpfish: Will do more testing

closed/make/Docs.gmk
lumpfish: Will change 'f' to f.

closed/make/Docs.gmk
lumpfish: The code is saying "find all files under DOCS_OUTPUTDIR/api which contain '/openj9' in their path (gets all files in an openj9xxxx sub-directory) and also contain .html ((probably) limits the files to those with an .html extension - this will include all those created by javadoc), and then substitute any occurrences of JAVADOC_BOTTOM in those files with the correct html footer for the code in question".  If there is an html (or other file) accidentally picked up by the grep which is not a javadoc created file I assume it will not contain the test JAVADOC_BOTTOM so the sed will be a noop.  My feeling is that keeping the filters simple will make it more easily understood and maintainable going forward.

closed/make/Docs.gmk
lumpfish: Well spotted.   Thanks.

closed/make/common/Modules.gmk
```diff
@@ -36,6 +36,17 @@ TOP_SRC_DIRS += \
...
+# openj9 specific modules for which javadoc is required
+DOCS_MODULES += \
+    openj9.cuda \
+    openj9.dataaccess \
+    openj9.dtfj \
+    openj9.gpu \
+    openj9.jvm \
+    openj9.sharedclasses \
+    openj9.traceformat \
+    #
```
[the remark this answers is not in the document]
lumpfish: The module openj9.dtfjview is intended to be used only by openj9.dtfj so javadoc is not required to be produced for it. This seemed to at least make it obvious which modules we are generating javadoc for.

closed/make/common/Modules.gmk
```diff
@@ -45,3 +56,4 @@ ifneq (,$(findstring create-main-targets-include,$(MAKECMDGOALS)))
 BUILD_OPENJ9_TOOLS     := $(shell $(MAKE) -C $(TOPDIR)/closed -f OpenJ9.gmk SPEC=$(SPEC) BOOT_JDK=$(BOOT_JDK) build-openj9-tools)
 GENERATE_J9JCL_SOURCES := $(shell $(MAKE) -C $(TOPDIR)/closed -f OpenJ9.gmk SPEC=$(SPEC) generate-j9jcl-sources)
 endif
+
```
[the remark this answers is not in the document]
lumpfish: Will remove

make/Docs.gmk
lumpfish: I tried very hard to get this to work by using the plugin point provided already.  Unfortunately some macros which need to be modified are changed after the include point, so they are overwritten.  Moving the include down a few lines works, but I'm not exactly comfortable with the 'solution'.  Perhaps it would be better to leave the existing include in place and an an extra include for a new file (e.g. Docs-post.gmk) in the new position.
If anyone can see how to make it work without this new include point I'd be very happy.

closed/make/Docs.gmk
keithc-ca: I suggest you at least replace the second `grep` with the extra predicate `-name '*.html'` in the `find` command: it's still clear and won't match arbitrary paths containing 'html'.

closed/make/common/Modules.gmk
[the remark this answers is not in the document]
keithc-ca: An explicit list means it will need to be updated when modules are added or removed.

The relationship between openj9.dtfj and openj9.dtfjview is the other way around: the latter requires the former. What's the problem with including javadoc for openj9.dtfjview?

make/Docs.gmk
keithc-ca: I'm happy to give it a try once you get past the build problem I noted. I'm hoping it just requires careful use of `=` instead of `:=` so the timing is right for macro expansion.

closed/make/common/Modules.gmk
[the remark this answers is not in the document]
pshipton: I don't think it is a problem to include openj9.dtfjview. The javadoc also includes other modules that don't contain content, such as jdk.jcmd, jdk.rmic. I've added a module comment, so we shouldn't get the copyright notice as the description any more.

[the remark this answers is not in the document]
lumpfish: OK, I'll include all the openj9.xxx modules then.

closed/make/Docs.gmk
lumpfish: All the targets run OK for me (make xxxx) except for docs-reference-api-javadoc which puts out some warnings like
`
Generating REFERENCE_API javadoc for 21 modules
javadoc: warning - bad source file: /root/openj9-openjdk-jdk11/build/linux-x86_64-normal-server-release/support/j9jcl_sources/java.base/share/classes/java/lang/LiveStackFrame.java
  file does not contain class java.lang.LiveStackFrame
  Please remove or make sure it appears in the correct subdirectory of the sourcepath.
javadoc: warning - bad source file: /root/openj9-openjdk-jdk11/build/linux-x86_64-normal-server-release/support/j9jcl_sources/java.base/share/classes/java/lang/StackFrameInfo.java
  file does not contain class java.lang.StackFrameInfo
  Please remove or make sure it appears in the correct subdirectory of the sourcepath.
`
for each of the files overridden in java.base, and then it hangs.  I don't think that is new with the changes I've made here but will investigate further.  I'm building using a docker image from the instructions here: https://www.eclipse.org/openj9/oj9_build.html, though my openj9 code needs bringing up to date.

lumpfish: I've pushed the changes I've made in response to the comments so far - please could you try the build again?
`make all`, `make docs` and `make docs-jdk-api-javadoc` should all generate the javadoc (the target for the generation in the api/index.html file, so if it already exists it will not regenerate).  Remove images/docs to force javadoc to run.
Note that at the moment modules `jdk.jfr, jdk.jstatd, jdk.management.jfr` are excluded from the build (and therefore also from the javadoc generation) - also openj9.dtfjview is still missing.

keithc-ca: I've got another build underway for `make all`. If that works it's a sign of some missing dependencies in the makefiles - it shouldn't matter what I build before trying to build the docs.

keithc-ca: This looks much like what happened in my first attempt yesterday:
```
find: ‘/home/keithc/space/jdk11/build/linux-x86_64-normal-server-release/support/src/src/jdk.management/share/classes/jdk.management/com/sun/management’: No such file or directory
find: ‘/home/keithc/space/jdk11/build/linux-x86_64-normal-server-release/support/src/src/jdk.management/share/classes/jdk.management/com/sun/management’: No such file or directory
/bin/bash: line 0: cd: /home/keithc/space/jdk11/build/linux-x86_64-normal-server-release/images/docs/api: No such file or directory
/home/keithc/space/jdk11/closed/make/Docs.gmk:117: recipe for target 'openj9-docs-set-javadoc-bottom' failed
make[3]: *** [openj9-docs-set-javadoc-bottom] Error 1
make/Main.gmk:364: recipe for target 'docs-jdk-api-javadoc' failed
make[2]: *** [docs-jdk-api-javadoc] Error 1
make[2]: *** Waiting for unfinished jobs....
```

keithc-ca: `make JOBS=1 all` succeeded: that suggest we're just missing some dependencies.

closed/make/Docs.gmk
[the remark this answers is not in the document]
lumpfish: Changed (now in closed/make/Docs-post.gmk).

closed/make/common/Modules.gmk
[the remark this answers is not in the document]
lumpfish: Changed to use a function.  Note that I've moved the openj9 javadoc generation code to a new Docs-post.gmk file so as not to move the original closed/make/Docs.gmk include in make/Docs.gmk (and added a new include for closed/make/Docs-post.gmk).

closed/make/Docs.gmk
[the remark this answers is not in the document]
lumpfish: Fixed (now in closed/make/Docs-post.gmk).

closed/make/Docs.gmk
lumpfish: Changed (now in closed/make/Docs-post.gmk).

closed/make/Docs.gmk
lumpfish: -name "*.html" added (now in closed/make/Docs-post.gmk).

make/Docs.gmk
lumpfish: I've moved the openj9 javadoc generation code to a new Docs-post.gmk file so as not to move the original closed/make/Docs.gmk include in make/Docs.gmk, and added a new include for closed/make/Docs-post.gmk.  This should at least reduce the chance of future merge conflicts.

closed/make/Docs.gmk
lumpfish: I think I've now fixed the dependencies and done much testing with "make JOBS=16 all" locally.

lumpfish: Looks like I spoke too soon - still getting some 'directory not found' errors running the build on other platforms.  Investigating......

lumpfish: These changes are now pushed to this PR, and the jdk and javadoc has been built successfully locally on x86-64_linux, ppc64le_linux, ppc64_aix, s390x_linux, x86-64_mac and x86-64_windows.

keithc-ca:
It's not clear to me that this new (better) strategy needs to introduce `Docs-post.gmk `: have you tried without that change?

lumpfish:
> 
> 
> It's not clear to me that this new (better) strategy needs to introduce `Docs-post.gmk `: have you tried without that change?

The issue is the overriding of the JAVADOC_BOTTOM macro so that the openj9 pages can have a different footer to the openjdk pages.  JAVADOC_BOTTOM is set (with `:=` ) after the provided plug-in point for a custom Docs.gmk.  The Docs-post.gmk file actually only needs the following two lines - the remaining code could all be in closed/make/Docs.mk (with an immediate assignment changed to a deferred assignment) , but my feeling was that it might be easier to follow if all the code related to the javadoc were in one file.
```
OPENJDK_JAVADOC_BOTTOM := $(JAVADOC_BOTTOM)
JAVADOC_BOTTOM := JAVADOC_BOTTOM
```
Of course these lines could be added to main/Docs.gmk directly instead.

lumpfish:
Looks like I should be able to prepend to GENERATED_SRC_DIRS to overcome the MODULE_SOURCE_PATH issue.  Investigating.....

lumpfish:
I now have a working version (pushed to this PR) which uses only the existing openjdk extension points.
1. The temporary directory containing the stubbed out files which are not supported on openj9 is added to GENERATED_SRC_DIRS in closed/make/common/Modules.gmk.
2. The openjdk javadoc generated .html page footers are replaced directly with the openj9 footers rather than using a placeholder mechanism.
Together these removed the need of the Docs-post.gmk file - the required changes are now added to the already existing closed/make/Docs.gmk.

lumpfish:
The unsupported com.sun.management files are to be stubbed out in the eclipse/openj9 source code.  This will remove the need to create temporary stub files at build times.  See https://github.com/eclipse/openj9/pull/7467.  I will rework this change.

lumpfish:
PR reworked to take advantage of the unsupported files now being stubbed out in the openj9 code base.
1. Change now uses the already existing plug-in point for closed/make/Docs.gmk
2. Solution now consists solely of adding the OpenJ9 modules to the list of modules for which javadoc is to be created, followed by replacing the html page footers for the OpenJ9 specific files.
3. The unsupported modules jdk.jfr, jdk.jstatd, jdk.management.jfr have been excluded from the build (and therefore also from the javadoc generation)

lumpfish:
I've successfully built jdk11 with these changes locally on ppc64_aix, x86-64_linux, ppc64le_linux, s390x_linux, x86-64_mac, x86-64_windows.

keithc-ca:
I've just returned from vacation: I expect to revisit this later this week.

I like that it only changes two files (both within the `closed` folder). Perhaps this should be squashed?

lumpfish:
Regarding squashing, yes we should squash.

closed/make/Docs.gmk
```diff
@@ -20,4 +20,77 @@
...
+    #
+JDK_GROUPS += OpenJ9
+
+##################################################################################
+# Add openj9 specific modules to the list of modules for which javadoc is required
+OPENJ9_MODULES := $(filter openj9.%,$(call FindAllModules))
+DOCS_MODULES += $(OPENJ9_MODULES)
+
+OPENJ9_JAVADOC_MARKER := $(SUPPORT_OUTPUTDIR)/docs/_javadoc_JDK_API_openj9.marker
+JDK_API_CUSTOM_TARGETS += $(OPENJ9_JAVADOC_MARKER)
```
keithc-ca: It doesn't appear that we also need to intercede in `docs-javase-api-javadoc` or `docs-reference-api-javadoc` targets: do you agree?

closed/make/Docs.gmk
```diff
@@ -20,4 +20,77 @@
...
+# Define a javadoc Group for the OpenJ9 specific modules
+# Module names are added to DOCS_MODULES in closed/make/common/Modules.gmk
+OpenJ9_GROUP_NAME := OpenJ9
+OpenJ9_GROUP_MODULES := openj9.*
+OpenJ9_GROUP_DESCRIPTION := \
+    Modules whose names start with {@code openj9.} contain APIs provided by the \
+    <a href="$(OPENJ9_JAVADOC_BASE_URL)">Eclipse OpenJ9</a> project. \
+    These APIs can only be used with a JDK which includes the OpenJ9 virtual machine. \
+    #
+JDK_GROUPS += OpenJ9
```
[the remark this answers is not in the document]
lumpfish: The macro names need to start with name name of the group (the value added to JDK_GROUPS) to which they relate - see the SetupApiDocsGenerationBody macro in https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/make/Docs.gmk.  Similar stanzas in that file set up JavaSE_GROUP_XXXX, JavaFX_GROUP_XXXX etc.

closed/make/Docs.gmk
lumpfish: I believe the `docs-javase-api-javadoc` or `docs-reference-api-javadoc` are there to produce the javadoc for just the javase packages and the reference implementation respectively, and therefore should not be altered to include the openj9 components.

closed/make/Docs.gmk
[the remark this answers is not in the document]
keithc-ca: I noticed that, but it should work as desired with:
```
JDK_GROUPS += OPENJ9
```
because there's also:
```
OPENJ9_GROUP_NAME := OpenJ9
```

closed/make/Docs.gmk
keithc-ca: I'll take that comment as agreement. This will need to change if we replace the javadoc for any of the 'javase' or 'reference' API classes (I had to check that things like `GcInfo` don't fall into that category).

closed/make/Docs.gmk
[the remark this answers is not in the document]
lumpfish: Yes, you're right.  I'll update the macros.

closed/make/Docs.gmk
```diff
@@ -20,4 +20,77 @@
...
+# '&' needs escaping in sed replacement text.
+#
+# {@docroot} is expanded by javadoc when the html files are created, so the matching
+# needs to cater for anything {@docroot} may have been substituted by - '../../../..'
+# or '../..' or '..' etc.
+
+# \(\.\./\)*\(\.\.\)* means zero  or more '../' strings followed by
+# zero or more '..' strings.
+
+JAVADOC_DOCROOT=\(\.\./\)*\(\.\.\)*
```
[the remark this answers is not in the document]
lumpfish: OK, will update the regex.

closed/make/Docs.gmk
```diff
@@ -20,4 +20,77 @@
...
+.SECONDEXPANSION:
+$(OPENJ9_JAVADOC_MARKER): $$(JDK_API_JAVADOC_TARGETS)
+	# Replace footer for openj9 modules, packages and classes
+	$(CD) $(DOCS_OUTPUTDIR)/api; find . -type f -name "*.html" | grep "/openj9" | xargs -I {} $(OPENJ9_SUBST_COMMAND) {}
+	# Replace footer for com.ibm packages and classes
+	$(CD) $(DOCS_OUTPUTDIR)/api; find . -type f -name "*.html" | grep "/com/ibm" | xargs -I {} $(OPENJ9_SUBST_COMMAND) {}
+	# Replace footer for the com.sun.management classes reimplemented in OpenJ9
+	$(CD) $(DOCS_OUTPUTDIR)/api; find . -type f -name "GarbageCollectionNotificationInfo.html" | xargs -I {} $(OPENJ9_SUBST_COMMAND) {}
+	$(CD) $(DOCS_OUTPUTDIR)/api; find . -type f -name "GcInfo.html" | xargs -I {} $(OPENJ9_SUBST_COMMAND) {}
+	$(CD) $(DOCS_OUTPUTDIR)/api; find . -type f -name "HotSpotDiagnosticMXBean.html" | xargs -I {} $(OPENJ9_SUBST_COMMAND) {}
```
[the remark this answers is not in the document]
lumpfish: The suggested change misses the module-summary.html files in the module root directories (e.g. openj9.cuda/module-summary.html).  This expression finds all the files:
```
	find . -type f \
        \( -path "*/openj9*/*.html" \
        -o -path "*/com/ibm/*/*.html" \
        -o -name GarbageCollectionNotificationInfo.html \
        -o -name GcInfo.html \
        -o -name HotSpotDiagnosticMXBean.html \)
```

lumpfish:
I've updated closed/make/Docs.gmk following the latest comments.

closed/make/Docs.gmk
```diff
@@ -20,4 +20,77 @@
...
+# {@docroot} is expanded by javadoc when the html files are created, so the matching
+# needs to cater for anything {@docroot} may have been substituted by - '../../../..'
+# or '../..' or '..' etc.
+
+# \(\.\./\)*\.\. means zero  or more '../' strings followed by '..'
+
+JAVADOC_DOCROOT=\(\.\./\)*\.\.
+JAVADOC_BOTTOM_REGEX = $(subst {@docroot},$(JAVADOC_DOCROOT),$(subst .,\.,$(JAVADOC_BOTTOM)))
+OPENJ9_JAVADOC_BOTTOM := $(subst &,\&,$(OPENJ9_JAVADOC_BOTTOM))
+OPENJ9_SUBST_COMMAND = sed -i 's|$(JAVADOC_BOTTOM_REGEX)|$(OPENJ9_JAVADOC_BOTTOM)|g'
```
keithc-ca: I have to wonder whether the pattern we need to replace is instead `(\.\./)*` because references in the generated html will be to files with a sequence of zero or more `../` at the beginning of those paths?

Again, please use `:=` instead of `=` in macro definitions (unless inappropriate).

closed/make/Docs.gmk
```diff
@@ -20,4 +20,77 @@
...
+##################################################################################
+# Replace the html footer in the generated javadoc for openj9 specific files.
+# Using the OPENJ9_JAVADOC_MARKER file as a target ensures the page footers are
+# only updated if the javadoc has been regenerated.
+
+$(OPENJ9_JAVADOC_MARKER): $(DOCS_OUTPUTDIR)/api/index.html
+	# Replace footer in the pages for
+	# 1. openj9 modules, packages and classes
+	# 2. com.ibm packages and classes
+	# 3. the com.sun.management classes reimplemented in OpenJ9
```
keithc-ca: Can you please move these comments out of the make script? Make will actually invoke the shell for each line.

closed/make/Docs.gmk
```diff
@@ -20,4 +20,77 @@
...
+	$(CD) $(DOCS_OUTPUTDIR)/api; \
+	find . -type f \
+        \( -path "*/openj9*/*.html" \
+        -o -path "*/com/ibm/*/*.html" \
+        -o -name GarbageCollectionNotificationInfo.html \
+        -o -name GcInfo.html \
+        -o -name HotSpotDiagnosticMXBean.html \) \
+    | xargs -I {} $(OPENJ9_SUBST_COMMAND) {}
```
keithc-ca: Please indent these lines to clarify that they are continuations of a single command.

closed/make/Docs.gmk
lumpfish: I've changed the JAVADOC_DOCROOT assignment to `JAVADOC_DOCROOT := \(\.\./\)*\.\.`.
`=` is required for the JAVADOC_BOTTOM_REGEX assignment (and therefore the OPENJ9_SUBST_COMMAND assignment) because JAVADOC_BOTTOM is not defined in make/Docs.gmk until after closed/make/Docs.gmk has been included.

lumpfish: The javadoc tool replaces @{docroot} with `..`, `../..`, `../../..` type references.  The regex `\(\.\./\)*` does not match unless `@{docroot}/` is replaced with the regex rather than just `@{docroot}`.  Since the intent is for the regex to replace the original `@{docroot}` then it seems best to leave it as it is.

closed/make/Docs.gmk
lumpfish: Moved

closed/make/Docs.gmk
lumpfish: Reformatted

closed/make/Docs.gmk
```diff
@@ -20,4 +20,77 @@
...
+$(OPENJ9_JAVADOC_MARKER): $(DOCS_OUTPUTDIR)/api/index.html
+	$(CD) $(DOCS_OUTPUTDIR)/api; \
+        find . -type f \
+            \( -path "*/openj9*/*.html" \
+            -o -path "*/com/ibm/*/*.html" \
+            -o -name GarbageCollectionNotificationInfo.html \
+            -o -name GcInfo.html \
+            -o -name HotSpotDiagnosticMXBean.html \) \
+            | xargs -I {} $(OPENJ9_SUBST_COMMAND) {}
+	# Update the timestamp on the marker file
```
keithc-ca: The leading whitespace for these lines should consist only of tabs.
Please also move the 'timestamp' comment before the rule.

lumpfish: OK.  I'll remove the timestamp comment if it can't be next to the command.  Hopefully it will be obvious.

closed/make/Docs.gmk
```diff
@@ -20,4 +20,77 @@
...
+# {@docroot} is expanded by javadoc when the html files are created, so the matching
+# needs to cater for anything {@docroot} may have been substituted by - '../../../..'
+# or '../..' or '..' etc.
+
+# \(\.\./\)*\.\. means zero  or more '../' strings followed by '..'
+
+JAVADOC_DOCROOT := \(\.\./\)*\.\.
+JAVADOC_BOTTOM_REGEX = $(subst {@docroot},$(JAVADOC_DOCROOT),$(subst .,\.,$(JAVADOC_BOTTOM)))
+OPENJ9_JAVADOC_BOTTOM := $(subst &,\&,$(OPENJ9_JAVADOC_BOTTOM))
+OPENJ9_SUBST_COMMAND = sed -i 's|$(JAVADOC_BOTTOM_REGEX)|$(OPENJ9_JAVADOC_BOTTOM)|g'
```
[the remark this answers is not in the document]
keithc-ca: Sorry, I missed your explanation that `=` is needed here. Perhaps a comment would help?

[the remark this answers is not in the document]
lumpfish: The remaining `=` need to be there - please see the conversation above - https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/212#discussion_r341546006

closed/make/Docs.gmk
lumpfish: Change committed.

keithc-ca:
This looks good. Can you please squash now?
I'd like to see you repeat this for jdk13 and jdknext so all can be updated together.
Some of the ideas here should be applied to ibmruntimes/openj9-openjdk-jdk8#342 as well.

keithc-ca:
Perhaps my declaration of success was premature: The list of files specific to OpenJ9 is incomplete, there are types such as `MonitorInfo` and `RuntimeMXBean` (those examples are in the `java.lang.management` package) that should be included as well.

lumpfish:
This PR originated from issues https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/20 and https://github.com/ibmruntimes/openj9-openjdk-jdk8/issues/21, (raised for jdk8 originally), which do not mention these additional openj9 specific changes.
I suggest we update those issues to document what the differences are.
Regarding this PR, I guess we can either commit what have so far and make further modificiations via another PR, or wait until they are all addressed here. @pshipton - do you have a preference?

pshipton:
I think we can wait until all the issues are addressed.

Are we generating the javadoc from the OpenJ9 implementation of the java.lang.management package or the OpenJDK implementation of this package? If we are using OpenJDK classes, in theory this is fine because these classes are part of the spec and have to match exactly. This is unlike com.ibm.lang.management which are extensions and have been modified by OpenJ9 so don't match exactly. If we are using OpenJ9 classes because they are what get compiled into the builds, I'm happy to update ibmruntimes/openj9-openjdk-jdk8#20 to include java.lang.managment into the list.

lumpfish:
For jdk11 (and later), the javadoc is generated according to the --module-source-path argument passed into the javadoc tool.  For openj9 builds, this puts support/j9jcl_sources ahead of the openjdk directories.  So currently, if an openjdk source file is overridden by an identically named openj9 file the javadoc is generated from the openj9 file.  This presumably is the correct behavior (notwithstanding the need to post process the html files to apply the correct page footers).
I will look into how to fix the footers for these additional files.

keithc-ca:
Don't forget there are also overrides in `closed/src`.

lumpfish:
Modified the code to update the page footers in:
1. .html pages with the same file stem as a j9jcl_sources .java file
2. .html pages with the same file stem as a closed/src .java file
3. .html pages beneath any docs/api directory with starts '/openj9'.
The code for No.2 is present although that directory actually does not seem to contain any public classes (the build does not create any javadoc for those files).
No.3 was necessary (even though the source files are present in j9jcl_sources) because multiple .html files can be produced from a single .java file (e.g. for enum classes), so just mapping .java source files to .html doc files is not sufficient.
I ended up writing the command to perform the substitution to a temporary file during the build because otherwise the expanded command exceeds the xargs command length limit of 255 on aix.

lumpfish:
@keithc-ca - please could you take another look?  I'll do jdk13 and head when we're happy with this since the changes will be similar.

closed/make/Docs.gmk
```diff
@@ -20,4 +20,93 @@
...
+			| sed 's|/[a-z][a-z]*/classes||g' \
+			| sed 's|\.java$$|\.html|g' \
+			| xargs -I {} sh $(OPENJ9_JAVADOC_SCRIPT) {}
+	$(CD) $(DOCS_OUTPUTDIR)/api; \
+		find . -type f -path "*/openj9*/*.html" \
+			| xargs -I {} sh $(OPENJ9_JAVADOC_SCRIPT) {}
+	$(CD) $(DOCS_OUTPUTDIR)/api; \
+		find . -type f -path "*/com/ibm*/*.html" \
+			| xargs -I {} sh $(OPENJ9_JAVADOC_SCRIPT) {}
+	$(TOUCH) $@
```
keithc-ca: * This should use macros BASH, FIND, SED and XARGS instead of sh, find, sed and xargs, respectively.
* The helper script can process more than one file at a time so xargs doesn't need to spawn as many shells; capturing it in a make variable makes it easier to read.
* SED can apply more than one substitution per invocation. The `g` suffix should not be used - each of the substitutions should occur at most once per line of input. We should be matching `/ibm/`, not `/ibm*/`.
* FIND can search for more than one pattern per invocation.
* The helper script can be removed once we're done using it.

The result is this:
```
OPENJ9_JAVADOC_SCRIPT_BODY = \
    for path in "$$@" ; do \
      if [ -f "$(DOCS_OUTPUTDIR)/api/$$path" ] ; then \
        $(CP) "$(DOCS_OUTPUTDIR)/api/$$path" "$(OPENJ9_JAVADOC_TEMP)" ; \
        $(OPENJ9_SUBST_COMMAND) < "$(OPENJ9_JAVADOC_TEMP)" > "$(DOCS_OUTPUTDIR)/api/$$path" ; \
        $(RM) "$(OPENJ9_JAVADOC_TEMP)" ; \
      fi ; \
    done \
    #

$(OPENJ9_JAVADOC_MARKER) : $(DOCS_OUTPUTDIR)/api/index.html
	$(ECHO) '$(OPENJ9_JAVADOC_SCRIPT_BODY)' >$(OPENJ9_JAVADOC_SCRIPT)
	$(CD) $(SUPPORT_OUTPUTDIR)/j9jcl_sources ; \
		$(FIND) . -type f -name '*.java' \
			| $(SED) -e 's|/[^/]\+/classes/|/|' -e 's|\.java$$|.html|' \
			| $(XARGS) $(BASH) $(OPENJ9_JAVADOC_SCRIPT)
	$(CD) $(TOPDIR)/closed/src ; \
		$(FIND) . -type f -name '*.java' \
			| $(SED) -e 's|/[^/]\+/classes/|/|' -e 's|\.java$$|.html|' \
			| $(XARGS) $(BASH) $(OPENJ9_JAVADOC_SCRIPT)
	$(CD) $(DOCS_OUTPUTDIR)/api ; \
		$(FIND) . -type f '(' -path '*/openj9*/*.html' ')' -o '(' -path '*/com/ibm/*.html' ')' \
			| $(XARGS) $(BASH) $(OPENJ9_JAVADOC_SCRIPT)
	$(RM) -f $(OPENJ9_JAVADOC_SCRIPT)
	$(TOUCH) $@
```

keithc-ca:
Thanks, @lumpfish. Please squash and I'll be happy to merge. A gentle reminder we'll want similar changes for jdk8, jdk13 and jdknext.

closed/make/Docs.gmk
```diff
@@ -20,4 +20,93 @@
...
+# {@docroot} is expanded by javadoc when the html files are created, so the matching
+# needs to cater for anything {@docroot} may have been substituted by - '../../../..'
+# or '../..' or '..' etc.
+
+# \(\.\./\)*\.\. means zero  or more '../' strings followed by '..'
+
+JAVADOC_DOCROOT := \(\.\./\)*\.\.
+JAVADOC_BOTTOM_REGEX = $(subst {@docroot},$(JAVADOC_DOCROOT),$(subst .,\.,$(JAVADOC_BOTTOM)))
+OPENJ9_JAVADOC_BOTTOM := $(subst &,\&,$(OPENJ9_JAVADOC_BOTTOM))
+OPENJ9_SUBST_COMMAND = sed '\''s|$(JAVADOC_BOTTOM_REGEX)|$(OPENJ9_JAVADOC_BOTTOM)|g'\''
```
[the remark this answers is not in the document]
lumpfish: Changed.

closed/make/Docs.gmk
lumpfish: Changed

lumpfish:
I've confirmed the build completes locally on ppc64_aix, x86-64_linux, ppc64le_linux, s390x_linux, x86-64_mac, x86-64_windows.

lumpfish:
Squashed.  Please merge if you're happy @keithc-ca .

keithc-ca:
Sorry, I should have also asked you to remove 'WIP:' from the title.

keithc-ca:
@pshipton Have all your comments been addressed to your satisfaction?

pshipton:
Is there a recent copy of the javadoc to look at? The last one provided is from Oct 24th.

keithc-ca:
I did a build and shared the javadoc with @pshipton.

pshipton:
The docs lgtm

keithc-ca:
@lumpfish Have you considered my suggestion (https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/156#discussion_r354508775) for an update here?

lumpfish:
Do you mean replacing
```
$(OPENJ9_JAVADOC_MARKER) : $(DOCS_OUTPUTDIR)/api/index.html
```
with
```
$(OPENJ9_JAVADOC_MARKER) : $(OPENJDK_JAVADOC_MARKER)
```
?
The use of a marker file by openjdk was introduced after jdk11 (it's used in jdk13 and head), so it's not available in jdk11.

keithc-ca:
The comment at [1] suggests otherwise (although the filename may differ).

The other difference it that you've captured the name of that file in a macro: I think it wise to do the same for Java 11.

[1] https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/make/Docs.gmk#L347

lumpfish:
Yes, I meant jdk11 uses the index.html file as the marker rather than a specific marker file which is 'touched' when the target runs.
Updated with the index.html file name assigned to a variable.

keithc-ca:
Jenkins compile aix,osx,plinux,xlinux,zlinux,win jdk11

keithc-ca:
Successful build at https://ci.eclipse.org/openj9/job/PullRequest-OpenJDK11/53/.
Please squash and I will merge.

lumpfish:
Squashed

PR #220, opened 2019-10-10
Update configure scripts for RISC-V (make)

ChengJin01:
The changes are to modify the scripts in the
make directory to enable RISC-V from the 
configuration perspective.

Issue: ibmruntimes#218

Signed-off-by: Cheng Jin <jincheng@ca.ibm.com>

### DIFF
--- a/make/autoconf/platform.m4
+++ b/make/autoconf/platform.m4
@@ -23,6 +23,10 @@
 # questions.
 #
 
+# ===========================================================================
+# (c) Copyright IBM Corp. 2019, 2019 All Rights Reserved
+# ===========================================================================
+
 # Support macro for PLATFORM_EXTRACT_TARGET_AND_BUILD.
 # Converts autoconf style CPU name to OpenJDK style, into
 # VAR_CPU, VAR_CPU_ARCH, VAR_CPU_BITS and VAR_CPU_ENDIAN.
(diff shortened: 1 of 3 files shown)


ChengJin01:
Already verified the compilation on Fedora_Linux/QEMU.

Reviewer: @keithc-ca , @pshipton , @DanHeidinga

make/autoconf/platform.m4
```diff
@@ -114,6 +118,12 @@ AC_DEFUN([PLATFORM_EXTRACT_VARS_FROM_CPU],
       VAR_CPU_BITS=64
       VAR_CPU_ENDIAN=little
       ;;
+    riscv64)
+      VAR_CPU=riscv64
+      VAR_CPU_ARCH=riscv64
+      VAR_CPU_BITS=64
+      VAR_CPU_ENDIAN=little
+      ;;
```
keithc-ca: Support for `riscv32` was added to the other files, but not here: why?

ChengJin01: `riscv32 ` was previously leftover option to check the configuration during the cross-compilation.
Given that there is no support on `riscv32 `in our plan, I removed it  just now in other files.

keithc-ca:
Jenkins compile plinux jdk11

PR #222, opened 2019-10-10
CMake: propagate CUDA enablement

keithc-ca:
* set CUDA_BIN_PATH so find_package(CUDA) uses the right version
* set OMR_OPT_CUDA=ON

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -493,8 +493,15 @@ ifeq (true,$(OPENJ9_ENABLE_CMAKE))
 
   CMAKE_ARGS += $(EXTRA_CMAKE_ARGS)
 
+  ifeq (true,$(OPENJ9_ENABLE_CUDA))
+    CMAKE_ARGS += -DOMR_OPT_CUDA=ON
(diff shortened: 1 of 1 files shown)


keithc-ca:
Without these changes, OMR tries to use `/usr/local/cuda` which may not exist or may be a version other than what the user specified via the `--with-cuda=` configure option.

FYI @dnakamura

closed/OpenJ9.gmk
```diff
@@ -493,8 +493,15 @@ ifeq (true,$(OPENJ9_ENABLE_CMAKE))
 
   CMAKE_ARGS += $(EXTRA_CMAKE_ARGS)
 
+  ifeq (true,$(OPENJ9_ENABLE_CUDA))
+  	CMAKE_ARGS += -DOMR_OPT_CUDA=ON
+    CMAKE_CUDA_ENV := CUDA_BIN_PATH="$(CUDA_HOME)"
```
dnakamura: Looks like mixed tabs & spaces here

keithc-ca: Yes - I'll fix that.

keithc-ca: Done.

pshipton:
jenkins compile xlinuxcm jdk11

PR #237, opened 2019-11-29
CMake: fix path when getting the redirector import lib on windows

dnakamura:


keithc-ca:
Jenkins test sanity win jdk11

PR #239, opened 2019-12-13
Add jitserver launcher in OpenJ9 builds

ashu-mehra:
If --enable-jitserver configuration option is set, then enable
building of jitserver launcher and include it in the jdk.

Signed-off-by: Ashutosh Mehra <mehra.ashutosh@ibm.com>

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -1,5 +1,5 @@
 # ===========================================================================
-# (c) Copyright IBM Corp. 2017, 2019 All Rights Reserved
+# (c) Copyright IBM Corp. 2017, 2020 All Rights Reserved
 # ===========================================================================
 # This code is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License version 2 only, as
@@ -170,6 +170,12 @@ openj9_add_jdk_lib = $(call openj9_add_jdk_basic, $(JDK_OUTPUTDIR)/lib $(call Fi
 openj9_add_jdk_special = \
 	$(foreach target, $(JDK_OUTPUTDIR)/$(OPENJ9_BIN_OR_LIB_DIR) $(call FindLibDirForModule, java.base), \
 		$(eval $(call openj9_add_jdk_rules, $(target)/$(strip $1), $(OPENJ9_VM_BUILD_DIR)/$(strip $2))))
+
+# jitserver
+
+ifeq (true,$(OPENJ9_ENABLE_JITSERVER))
+  $(call openj9_add_jdk_basic, $(JDK_OUTPUTDIR)/bin $(call FindExecutableDirForModule, java.base), jitserver$(EXE_SUFFIX))
+endif
 
 # redirector
 
@@ -361,6 +367,13 @@ ifeq (true,$(OPENJ9_ENABLE_DDR))
 else
   FEATURE_SED_SCRIPT += $(call SedDisable,opt_useOmrDdr)
   SPEC_SED_SCRIPT    += $(call SedDisable,module_ddr)
+endif
+
+# Adjust JITServer enablement flags.
+ifeq (true,$(OPENJ9_ENABLE_JITSERVER))
+  FEATURE_SED_SCRIPT += $(call SedEnable,build_jitserver)
+else
+  FEATURE_SED_SCRIPT += $(call SedDisable,build_jitserver)
 endif
 
 # Disable windows rebase.


ashu-mehra:
fyi @keithc-ca

ashu-mehra:
@pshipton @keithc-ca can you please review this PR. It is required for https://github.com/eclipse/openj9/pull/7940.

ashu-mehra:
@pshipton @keithc-ca can you please review this PR.

closed/OpenJ9.gmk
```diff
@@ -171,6 +171,12 @@ openj9_add_jdk_special = \
 	$(foreach target, $(JDK_OUTPUTDIR)/$(OPENJ9_BIN_OR_LIB_DIR) $(call FindLibDirForModule, java.base), \
 		$(eval $(call openj9_add_jdk_rules, $(target)/$(strip $1), $(OPENJ9_VM_BUILD_DIR)/$(strip $2))))
 
+# jitserver
+
+ifeq (true,$(OPENJ9_ENABLE_JITSERVER))
+$(call openj9_add_jdk_basic, $(JDK_OUTPUTDIR)/bin $(call FindExecutableDirForModule, java.base), jitserver)
```
keithc-ca: This should say `jitserver$(EXE_SUFFIX)` instead of just `jitserver`.
A copy should also be placed in `$(call FindExecutableDirForModule, java.base)`.

keithc-ca:
The resulting executable is not stripped of debug information like it is for Java 8.

ashu-mehra:
> The resulting executable is not stripped of debug information like it is for Java 8.

I checked the build scripts but I am not sure how to do that for Java 11. @keithc-ca you have any idea? I can add command in `closed/OpenJ9.gmk` to strip the debug info if that is ok?

keithc-ca:
> > The resulting executable is not stripped of debug information like it is for Java 8.
> 
> I checked the build scripts but I am not sure how to do that for Java 11. @keithc-ca you have any idea? I can add command in `closed/OpenJ9.gmk` to strip the debug info if that is ok?

This is another reason for my [comment](https://github.com/eclipse/openj9/issues/7800#issuecomment-565053914); I'll have a look to see if I can find something we can reuse to strip the launcher.

keithc-ca:
I think the right answer is to use `objcopy` in the script that links `jitserver` like is [done for shared libraries in OMR](https://github.com/eclipse/omr/blob/4c52666cdb2d99bc4d54be59ea8a726120b1079a/omrmakefiles/rules.linux.mk#L423-L428); perhaps as part of https://github.com/eclipse/openj9/pull/7940.

ashu-mehra:
OpenJ9 does similar stuff in its makefile for [dlls](https://github.com/eclipse/openj9/blob/14d1f89b9ee42deab3690efe11e0deb4e81c8096/runtime/makelib/targets.mk.linux.inc.ftl#L46). I guess we can update the [exe_target_rule](https://github.com/eclipse/openj9/blob/14d1f89b9ee42deab3690efe11e0deb4e81c8096/runtime/makelib/targets.mk.linux.inc.ftl#L55) in the same way. But it wouldn't apply to cmake build system. Not sure if we strip out debug info in cmake builds, do we?

keithc-ca:
The cmake build should be doing the same things (or it's broken).

ashu-mehra:
It looks like with cmake the OpenJ9 libraries are not getting stripped of debug info when they are compiled (i.e. libraries in `build/linux-x86_64-normal-server-release/vm/runtime`), but in the final jdk (i.e. `build/linux-x86_64-normal-server-release/images/j2sdk-image`) the libraries are stripped of the debug info.
I am guessing the openjdk make files are doing that when the libraries are copied from the `vm` dir to `images` dir.

ashu-mehra:
So if we update the OpenJ9 makefiles, the cmake builds would still have `jitserver` exe with debug info.

keithc-ca:
> So if we update the OpenJ9 makefiles, the cmake builds would still have `jitserver` exe with debug info.

I'm fine with deferring the changes to strip debug info from executables to a separate PR (but it should be for both styles of builds). Please open an issue if you want to proceed that way.

ashu-mehra:
> I'm fine with deferring the changes to strip debug info from executables to a separate PR (but it should be for both styles of builds). Please open an issue if you want to proceed that way.

Ok, I will open an issue in OpenJ9 to track it.

closed/OpenJ9.gmk
ashu-mehra: Fixed.

keithc-ca:
> > I'm fine with deferring the changes to strip debug info from executables to a separate PR (but it should be for both styles of builds). Please open an issue if you want to proceed that way.
> 
> Ok, I will open an issue in OpenJ9 to track it.

Can you please share a link to that issue? I'd like to see it created before this is merged.

ashu-mehra:
> Can you please share a link to that issue? I'd like to see it created before this is merged.

Here is the issue: https://github.com/eclipse/openj9/issues/8219

ashu-mehra:
I guess this PR can now be merged.

PR #241, opened 2020-01-06
Remove the check on a specific OpenSSL API for JITServer

a7ehuo:
With the change in eclipse/openj9#8145, `OpenSSL` symbols are
dynamically loaded at the runtime. During the build time,
there is no longer a dependency on the installed `OpenSSL`
version on the build machines. Removed the code that
checks `SSL_CTX_set_ecdh_auto` during config.

Signed-off-by: Annabelle Huo <Annabelle.Huo@ibm.com>


keithc-ca:
Jenkins compile plinux jdk11

PR #242, opened 2020-01-08
Fix crash in NativeCrypto_CBCUpdate openssl support

shanchao95:
* remove finalize
* add a Cleaner

Fixes https://github.com/eclipse/openj9/issues/8181

Signed-off-by: Chao Shan <chao.shan@ibm.com>

### DIFF
--- a/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeCipherBlockChaining.java
+++ b/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeCipherBlockChaining.java
@@ -24,7 +24,7 @@
  */
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2019 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2020 All Rights Reserved
  * ===========================================================================
  */
 
@@ -36,6 +36,8 @@ import java.util.ArrayDeque;
 import com.sun.crypto.provider.AESCrypt;
 
 import jdk.crypto.jniprovider.NativeCrypto;
+import jdk.internal.ref.CleanerFactory;
+import java.lang.ref.Cleaner;
 
 /**
  * This class represents the native implementation of ciphers in
@@ -52,13 +54,15 @@ class NativeCipherBlockChaining extends FeedbackCipher  {
     protected static long[] contexts;
     protected static ArrayDeque<Integer> avStack = new ArrayDeque<Integer>(numContexts);
 
-    private static NativeCrypto nativeCrypto;
+    private static final NativeCrypto nativeCrypto;
+    private static final Cleaner contextCleaner;
 
     /*
      * Initialize the CBC context.
      */
     static {
         nativeCrypto = NativeCrypto.getNativeCrypto();
+        contextCleaner = CleanerFactory.cleaner();
         contexts = new long[numContexts];
 
         for (int i = 0; i < numContexts; i++) {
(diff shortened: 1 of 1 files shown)


closed/src/java.base/share/classes/com/sun/crypto/provider/NativeCipherBlockChaining.java
```diff
@@ -72,6 +76,33 @@
...
+            this.ctxIndx = ctxIndx;
+            this.nativeContext = nativeContext;
+        }
+
+        @Override
+        public void run() {
+            /*
+            * Release the CBC context.
+            */
+            if(ctxIndx == -1) {
```
pshipton: This isn't going to work. As currently written, ctxIndx here is a copy of the value in the NativeCipherBlockChaining instance, and doesn't get updated.

ashbm5: @pshipton I don't understand your comment - that's the intention that ctxIndx will be a copy of the value in the NativeCipherBlockChaining instance. ctxIndx is assigned to the NativeCipherBlockChaining instance in the constructor (in the call getContext()) and doesn't change for the lifetime of the instance, note that only getContext() changes ctxIndx. At cleaning time, if the ctxIndx value is -1 a special release should be taken. otherwise the value should be returned to the global object avStack.

pshipton: Ok, I see it now.
For consistency, please add a space, it should if `if (`

pshipton:
The commit should have a better title, like "Fix crash in NativeCrypto_CBCUpdate openssl support". Something that is meaningful on its own when looking at the titles of commits.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeCipherBlockChaining.java
```diff
@@ -72,6 +76,33 @@
         }
     }
 
+    private static final class CBCCleanerRunnable implements Runnable {
+        private int ctxIndx;
+        private long nativeContext;
```
keithc-ca: These fields should be `final`.

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeCipherBlockChaining.java
```diff
@@ -72,6 +76,33 @@
...
+        public CBCCleanerRunnable(int ctxIndx, long nativeContext) {
+            this.ctxIndx = ctxIndx;
+            this.nativeContext = nativeContext;
+        }
+
+        @Override
+        public void run() {
+            /*
+            * Release the CBC context.
+            */
```
keithc-ca: nit: please fix indentation (so `*`s line up like other comments).

keithc-ca:
When we've arrived at the final version, it should be propagated to jdk8, jdk13, jdk14 and jdk(next).

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeCipherBlockChaining.java
```diff
@@ -72,6 +76,33 @@
...
+        public CBCCleanerRunnable(int ctxIndx, long nativeContext) {
+            this.ctxIndx = ctxIndx;
+            this.nativeContext = nativeContext;
+        }
+
+        @Override
+        public void run() {
+    /*
+     * Release the CBC context.
+     */
```
keithc-ca: This should be indented 8 more spaces (to line up with the code it describes).

shanchao95:
> When we've arrived at the final version, it should be propagated to jdk8, jdk13, jdk14 and jdk(next).

keithc-ca:
Jenkins test sanity,extended plinux jdk11

jdekonin:
Sorry, the extended functional failure was likely me attempting to clean up dead processes on machine. 

Jenkins test extended plinux jdk11

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeCipherBlockChaining.java
```diff
@@ -72,6 +76,33 @@
...
+            this.nativeContext = nativeContext;
+        }
+
+        @Override
+        public void run() {
+            /*
+             * Release the CBC context.
+             */
+            if (ctxIndx == -1) {
+                long ret = nativeCrypto.DestroyContext(nativeContext);
```
keithc-ca: Previously, this call occurred within a `synchronized (NativeCipherBlockChaining.class)` block: Do we *know* that it is safe for multiple threads to call `nativeCrypto.DestroyContext()` concurrently?

shanchao95: Thanks, i will change to both `avStack.push(ctxIndx)` and `nativeCrypto.DestroyContext(nativeContext)` within a synchronized (NativeCipherBlockChaining.class) for safety.

keithc-ca:
Jenkins test sanity,extended plinux jdk11

keithc-ca:
Please prepare equivalent changes for jdk8, jdk13, jdk14 and jdk(next).

PR #243, opened 2020-01-08
Remove variables relating to NO_USE_MINGW

keithc-ca:
* they no longer have any effect


pshipton:
jenkins compile win jdk11

PR #249, opened 2020-01-14
Only build jzos stubs on non-z/OS platforms

pshipton:
Use the com.ibm.jzos module on z/OS.

Depends on https://github.com/eclipse/openj9/pull/8299


pshipton:
Once this looks good, I'll create the changes for the other repos.

keithc-ca:
> Once this looks good, I'll create the changes for the other repos.

I think we're ready for those other PRs.

PR #250, opened 2020-01-15
Merge jdk-11.0.6+10

andrew-m-leonard:


lumpfish:
Overriding checks because changes merged from openjdk do not conform to the check rules.

PR #254, opened 2020-01-17
Stop staging jcl and sourcetools folders

keithc-ca:
* staged jcl folder was not used
* build tools directly from unstaged sourcetools folder
* add missing PHONY declaration for clean-openj9-thirdparty-binaries
* don't assume CUDA is disabled by default
* remove unnecessary definition of OPENJ9_BUILD

Enabled by eclipse/openj9#8246: improves build performance.


keithc-ca:
Jenkins test sanity zlinux,win jdk11

PR #256, opened 2020-01-20
Improve integration of OpenJ9 JCL code generation

keithc-ca:
Generation of J9JCL code was connected to the build system in an unusual manner that triggered the work at least twice and hid all diagnostics from the build log. This adds a new target that is a prerequisite of `create-main-targets-include` which operates as a normal make rule.


pshipton:
jenkins compile plinux,win jdk11

pshipton:
jenkins compile xlinuxcm jdk11

PR #261, opened 2020-01-31
Sign bundled crypto libraries

keithc-ca:
When libcrypto.dylib or libssl.dylib are bundled in a jdk on OSX, they must also be signed.

* define CodesignFile in custom-spec.gmk and use in OpenJ9.gmk, Copy-java.base.gmk
* change AIX search order for libcrypto to match jdk8

Issue: eclipse/openj9#8389


pshipton:
jenkins test sanity osx,win,xlinuxcm,plinux,aix jdk11

pshipton:
The cmake sanity test was unexpected, I don't expect the failures are related to this PR.

PR #269, opened 2020-02-25
Add new flag J9VM_OPT_JITSERVER to enable JITServer

ashu-mehra:
The macro used for building JITServer in OpenJ9 is changed from
JITSERVER_SUPPORT to J9VM_OPT_JITSERVER. The makefile OpenJ9.mk
needs to be updated to set the correct env variable.
~~Older env variable JITSERVER_SUPPORT would be kept until changes
in OpenJ9 are not done to avoid any build breaks.~~

Signed-off-by: Ashutosh Mehra <mehra.ashutosh@ibm.com>

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -219,9 +219,9 @@ endif
 
 # Adjust JITServer enablement flags.
 ifeq (true,$(OPENJ9_ENABLE_JITSERVER))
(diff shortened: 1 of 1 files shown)


ashu-mehra:
See https://github.com/eclipse/openj9/pull/8656 for related OpenJ9 change.

PR #273, opened 2020-03-11
Always copy OpenJ9 debug information

keithc-ca:
Part of the solution to eclipse/openj9#8821.


pshipton:
jenkins compile zlinux jdk11

pshipton:
jenkins compile zlinux jdk11

PR #279, opened 2020-03-19
Set environment variable MALLOCOPTIONS=multiheap on AIX

pshipton:
The previous setting `MALLOCOPTIONS=multiheap,buckets` caused
Issue eclipse/openj9#8842


pshipton:
I think the testing in https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/389 is good enough.

Issue #231, opened 2019-11-08
Enable DDR for z/OS


pshipton:
It was enabled and then reverted because it caused some build errors. See https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/230

keithc-ca:
It seems we have a problem on z/OS similar to https://github.com/eclipse/omr/issues/3354 on OSX.

PR #284, opened 2020-03-27
Enable DDR on z/OS by default

keithc-ca:
Depends on eclipse/openj9#9011
Fixes: #231


pshipton:
This should go into the head stream (openjdk) as well, so it's set for the future.

PR #288, opened 2020-04-06
Use CloseHandle() to destroy a mutex on Windows

keithc-ca:
As suggested in https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/381#discussion_r404423466.


pshipton:
I expect the testing done in https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/396 is good enough for this identical change.

PR #291, opened 2020-04-08
Fix J9DDR compile commands

keithc-ca:
Compiling DDR_VM and the related generated code must use `--upgrade-module-path` to use current classes (e.g. from openj9.dtfj) else builds may fail (by using older, incompatible classes from the boot JDK).

### DIFF
--- a/closed/DDR.gmk
+++ b/closed/DDR.gmk
@@ -185,6 +185,7 @@ DDR_ALIAS_FILES := \
 # Compile the Java sources.
 $(eval $(call SetupJavaCompilation,BUILD_J9DDR_MAIN_CLASSES, \
 	SETUP := GENERATE_JDKBYTECODE, \
+	ADD_JAVAC_FLAGS := --upgrade-module-path $(JDK_OUTPUTDIR)/modules, \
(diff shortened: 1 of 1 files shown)


pshipton:
jenkins test extended zlinux jdk11

pshipton:
Does this fix a known problem (an existing issue), or it's a new problem discovered?

keithc-ca:
> Does this fix a known problem (an existing issue), or it's a new problem discovered?

I discovered this working on changes to have `jdmpview` warn about truncated core files. To that end, I added to `com.ibm.dtfj.image.Image`:
```
	/**
	 * Is this image truncated (i.e. incomplete)?
	 */
	public default boolean isTruncated() {
		return false;
	}
```
Without this, `com.ibm.j9ddr.view.dtfj.image.J9DDRImage`, which implements the above interface, fails to compile because the `@Override` annotation can't be connected to the new method (the interface `com.ibm.dtfj.image.Image` resolves to the one exported by `openj9.dtfj` in the boot jdk - if it is an OpenJ9 implementation).

PR #299, opened 2020-04-28
Rename the com.ibm.jzos module to ibm.jzos

pshipton:
Depends on https://github.com/eclipse/openj9/pull/9380


keithc-ca:
Tested in eclipse/openj9#9380.

PR #300, opened 2020-04-29
Add permissions to create system Thread for openj9.cuda & gpu

pshipton:
Required for https://github.com/eclipse/openj9/pull/9390


keithc-ca:
Tested in eclipse/openj9#9390.

PR #304, opened 2020-05-05
WIP: Check the correct message in ERR_RECEIVED when JITServer is enabled

a7ehuo:
`InheritIO.sh` fails the test if `stderr.txt` is not equal to `"exit value: 0"`. With JITServer is enabled, the first message in `stderr.txt` is `"JITServer is currently a technology preview. Its use is not yet supported"` before the second message `"exit value: 0"`. Updated `ERR_RECEIVED` to retrieve the correct message when JITServer is enabled.

Related to eclipse/openj9#9169

Signed-off-by: Annabelle Huo <Annabelle.Huo@ibm.com>

### DIFF
--- a/test/jdk/java/lang/ProcessBuilder/InheritIO/InheritIO.sh
+++ b/test/jdk/java/lang/ProcessBuilder/InheritIO/InheritIO.sh
@@ -74,7 +74,7 @@ do
     fi
 
     ERR_EXPECTED='exit value: 0'
-    ERR_RECEIVED=`cat stderr.txt`
+    ERR_RECEIVED=`tail -n 1 stderr.txt`
     if [ "x${ERR_RECEIVED}" != "x${ERR_EXPECTED}" ]; then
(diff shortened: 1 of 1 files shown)


pshipton:
We don't want this type of change in the OpenJDK extensions project. The number of patches to OpenJDK code should remain minimal and ideally the patches contributed back to OpenJDK. Every patch has the potential for causing a merge conflict when accepting new OpenJDK code, which then has to be fixed manually. Instead, I suggest adding an option to suppress the JITServer warning message, and run the tests with this option enabled.

a7ehuo:
> I suggest adding an option to suppress the JITServer warning message, and run the tests with this option enabled.

Sounds good. I'll look into it. Thanks @pshipton!

a7ehuo:
The issue will be addressed by eclipse/openj9#9483 and AdoptOpenJDK/openjdk-tests#1777.

PR #305, opened 2020-05-11
Added special case for EngineAEADDec in NativeChaCha20Cipher.engineUpdate

samasri:
Fixing [OpenJ9 Issue #9429](https://github.com/eclipse/openj9/issues/9429).

I added a special case in the ChaCha20 update function: when the size returned by the `doUpdate` is zero, the function must return a new zero-sized array since the input was only buffered and nothing was encrypted.

Signed-off-by: Samer AL Masri <samasri@ibm.com>

### DIFF
--- a/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
+++ b/closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
@@ -636,7 +636,12 @@ abstract class NativeChaCha20Cipher extends CipherSpi {
     protected byte[] engineUpdate(byte[] in, int inOfs, int inLen) {
         byte[] out = new byte[inLen];
         try {
-            engine.doUpdate(in, inOfs, inLen, out, 0);
+            int size = engine.doUpdate(in, inOfs, inLen, out, 0);
+            // Special case for EngineAEADDec, doUpdate only buffers the input
+            // So the output array must be empty since no encryption happened yet
+            if (size == 0) {
+                return new byte[0];
+            }
         } catch (ShortBufferException | KeyException exc) {
             throw new RuntimeException(exc);
         }


closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
```diff
@@ -636,7 +636,10 @@ private void checkKeyAndNonce(byte[] newKeyBytes, byte[] newNonce)
     protected byte[] engineUpdate(byte[] in, int inOfs, int inLen) {
         byte[] out = new byte[inLen];
         try {
-            engine.doUpdate(in, inOfs, inLen, out, 0);
+            int size = engine.doUpdate(in, inOfs, inLen, out, 0);
+            // Special case for EngineAEADDec, doUpdate only buffers the input
+            // So the output array must be empty since no encryption happened yet
+            if(size == 0) return new byte[0];
```
pshipton, line 642: Pls add a space after `if`.
We'll need PRs for jdk14 and jdk as well.

pshipton:
jenkins compile osx jdk11

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
samasri: Done

pshipton:
https://ci.eclipse.org/openj9/job/Build_JDK11_x86-64_mac_Personal/492/

pshipton:
Confirmed the excluded test com/sun/crypto/provider/Cipher/ChaCha20/OutputSizeTest.java passes with this change in a grinder.  https://ci.eclipse.org/openj9/view/Test/job/Grinder/878

Better check other problems aren't introduced.

jenkins test sanity.openjdk osx jdk11

pshipton:
The previous test run passed. https://ci.eclipse.org/openj9/job/Test_openjdk11_j9_sanity.openjdk_x86-64_mac_Personal/2/

closed/src/java.base/share/classes/com/sun/crypto/provider/NativeChaCha20Cipher.java
keithc-ca: Don't put `if` and body on a single line, instead write:
```
if (size == 0) {
    return new byte[0];
}
```

samasri: Done

PR #306, opened 2020-05-12
Ignore warnings only for certain modules

keithc-ca:
* inspired by ibmruntimes/openj9-openjdk-jdk#202


keithc-ca:
Jenkins compile xlinux jdk11

PR #308, opened 2020-05-21
Default to CMake builds on osx

dnakamura:

### DIFF
--- a/closed/autoconf/custom-hook.m4
+++ b/closed/autoconf/custom-hook.m4
@@ -57,21 +57,26 @@ AC_DEFUN([OPENJ9_CONFIGURE_CMAKE],
       if test "x$with_cmake" == xyes -o "x$with_cmake" == x ; then
         with_cmake=cmake
       fi
-      if test "x$with_cmake" != xno ; then
-        if AS_EXECUTABLE_P(["$with_cmake"]) ; then
-          CMAKE="$with_cmake"
-        else
-          BASIC_REQUIRE_PROGS([CMAKE], [$with_cmake])
-        fi
-        with_cmake=yes
-      fi
     ],
-    [with_cmake=no])
-  if test "$with_cmake" == yes ; then
-    OPENJ9_ENABLE_CMAKE=true
-  else
+    [
+      if test "x$OPENJDK_BUILD_OS" = xmacosx ; then
+        with_cmake=cmake
+      else
+        with_cmake=no
+      fi
+    ])
+  # at this point with_cmake should either be no, or the name of the cmake command
+  if test "x$with_cmake" = xno ; then
     OPENJ9_ENABLE_CMAKE=false
+  else
+    OPENJ9_ENABLE_CMAKE=true
+    if AS_EXECUTABLE_P(["$with_cmake"]) ; then
+      CMAKE="$with_cmake"
+    else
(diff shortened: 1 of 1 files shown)


dnakamura:
Marked as WIP as we are not yet ready to switch over the default
@keithc-ca  can you take a look at this now, to minimize delay when we are ready to flip the switch?

closed/autoconf/custom-hook.m4
```diff
@@ -53,22 +53,22 @@ AC_DEFUN_ONCE([CUSTOM_EARLY_HOOK],
...
+        with_cmake=no
       fi
-    ],
-    [with_cmake=no])
-  if test "$with_cmake" == yes ; then
+    ])
+
+  if test "$with_cmake" != no ; then
     OPENJ9_ENABLE_CMAKE=true
+    BASIC_PATH_PROGS(CMAKE, cmake)
```
keithc-ca, line 68: This used to support `--with-cmake=alternate-cmake-version`: please restore that capability.

dnakamura: This can still be achived by setting the `CMAKE` environment variable/option on the command line. This is more in line with how you would specify alternate versions of other tools. eg setting the `CC` variable

keithc-ca: If the configure option were `--enable-cmake` I would agree with you, but the convention is that `--with-XXX` options accept values (see also `--with-openj9-cc` and  `--with-openj9-cxx`, etc.). Please don't break existing behavior just to change the default for one platform.

closed/autoconf/custom-hook.m4
```diff
@@ -57,21 +57,26 @@ AC_DEFUN([OPENJ9_CONFIGURE_CMAKE],
...
-  fi
+    [
+      if test "x$OPENJDK_BUILD_OS" = xmacosx ; then
+        with_cmake=cmake
+      else
+        with_cmake=no
+      fi
+    ])
+    # at this wpoint with_cmake should either be no, or the name of the cmake cmd
+    if test "$with_cmake" != no ; then
```
keithc-ca, line 69: typo: wpoint -> point
I suggest this `if` statement would be easier to follow if inverted (so `if` and `else` are closer to each other).

closed/autoconf/custom-hook.m4
```diff
@@ -57,21 +57,26 @@ AC_DEFUN([OPENJ9_CONFIGURE_CMAKE],
...
+      OPENJ9_ENABLE_CMAKE=true
+      if AS_EXECUTABLE_P(["$with_cmake"]) ; then
+        CMAKE="$with_cmake"
+      else
+        BASIC_REQUIRE_PROGS([CMAKE], [$with_cmake])
+      fi
+    else
+      OPENJ9_ENABLE_CMAKE=false
+    fi
```
keithc-ca, line 78: This is indented by two spaces too many.

keithc-ca:
Jenkins compile osx,plinux jdk11

pshipton:
The PR build was lost in the jenkins update this morning. Restarting it.

Jenkins compile osx,plinux jdk11

pshipton:
More accurately, the original builds did pass, but didn't upload properly, and look like failures here. My new PR build request didn't get reflected here, but is running https://ci.eclipse.org/openj9/job/PullRequest-OpenJDK11/101/

PR #316, opened 2020-06-17
Fix handling of reference repos

keithc-ca:
Avoid unexpected reuse of reference repos: if `-openj9-reference=` or `-omr-reference=` are specified, but not both, the one reference may be used by both clones: This fixes that.

### DIFF
--- a/closed/get_j9_source.sh
+++ b/closed/get_j9_source.sh
@@ -8,7 +8,7 @@
 # under the terms of the GNU General Public License version 2 only, as
 # published by the Free Software Foundation.
 #
-# IBM designates this particular file as subject to the "Classpath" exception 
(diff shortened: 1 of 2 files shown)


keithc-ca:
Looks like similar changes will be required for all extension repos: coming up...

pshipton:
https://github.com/ibmruntimes/openj9-openjdk-jdk14/pull/58 hasn't been merged yet.

keithc-ca:
I don't yet see a change for jdknext: @AdamBrousseau will you be putting that together?

pshipton:
jdknext change is https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/212

pshipton:
Just merged 30 min ago.

PR #318, opened 2020-06-19
Fix deserializeLambda being generated with bad string encoding

mikezhang1234567890:
When javac creates the `deserializeLambda` method, the `externalize` method is
called on the class names needed in order to swap the `.` and `/` characters.
This is returned as a byte array, which is turned into a string by calling the
String constructor. However, the bytes are encoded in the JVM's UTF encoding,
which can result in the String constructor making a bad string if the platform
default encoding has different byte values for characters.
This commit uses the `Names.fromUtf` method also used in ClassWriter.java for
dealing with the externalized class names instead of the String constructor.

Signed-off-by: Mike Zhang <mike.h.zhang@ibm.com>

Fixes https://github.com/eclipse/openj9/issues/9168

### DIFF
--- a/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LambdaToMethod.java
+++ b/src/jdk.compiler/share/classes/com/sun/tools/javac/comp/LambdaToMethod.java
@@ -23,6 +23,12 @@
  * questions.
  */
 
+ /*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2020, 2020 All Rights Reserved.
+ * ===========================================================================
+ */
+
 package com.sun.tools.javac.comp;
 
 import com.sun.tools.javac.tree.*;
@@ -2464,7 +2470,8 @@ public class LambdaToMethod extends TreeTranslator {
 
         @Override
         protected void append(byte[] ba) {
-            sb.append(new String(ba));
+            Name name = names.fromUtf(ba);
+            sb.append(name.toString());
         }
 
         @Override


pshipton:
@mikezhang1234567890 the file need to have an IBM copyright added, after the existing license.
```
/*
 * (c) Copyright IBM Corp. 2020, 2020 All Rights Reserved.
 */
```

pshipton:
jenkins test sanity zlinux jdk11

pshipton:
https://ci.eclipse.org/openj9/view/Pull%20Requests/job/PullRequest-OpenJDK11/103/

mikezhang1234567890:
Which other extension repos does this need to be delivered to? 8, 14, 15, and head?

pshipton:
>  8, 14, 15, and head?

yup. @DanHeidinga are you ok with delivering this fix to the 0.21 branches as well?

DanHeidinga:
> yup. @DanHeidinga are you ok with delivering this fix to the 0.21 branches as well?

I'd like us to start the conversation on upstreaming this before we put it in the release.  I don't expect any issue with getting it accepted at OpenJDK but would prefer not to carry patches that aren't already acknowledged as bug fixes upstream.

We don't have to wait for it to be backported upstream, acknowledgement and targeting to the `jdk` stream would be sufficient.

pshipton:
@andrew-m-leonard can you please start the conversation with OpenJDK. I've attached a test case that shows the problem.

My machine is using UTF8 so the â character I put in the class name translates to the bytes 0xc3 0xa2 in the TestSerial.java file on disk. The byte representation is important, as they need to represent legal characters to javac in the Charset we're compiling in (ISO-8859-1 in this case).

Modify the platform default Charset when running javac, or compile on a machine that isn't using UTF8, ideally ISO-8859-1 for the purposes of this test.
javac -J-Dfile.encoding=ISO-8859-1 TestSerial.java

Run the test normally using the default UTF8 Charset.
`java TestSerial`
results in `IllegalArgumentException: Invalid lambda deserialization`

[TestSerial.zip](https://github.com/ibmruntimes/openj9-openjdk-jdk11/files/4815190/TestSerial.zip)

andrew-m-leonard:
yes that's a bug, i'll raise bug and try and fix upstream, remember will get fixed in jdk(next) first before being backported..

andrew-m-leonard:
https://bugs.openjdk.java.net/browse/JDK-8248231

andrew-m-leonard:
@pshipton hit a problem trying to create a Jtreg testcase for this, it seems when run under the openjdk test framework, the javac FileManager has problems with the file path ABC??.class:
```
 stderr: [/home/admin/mydocker/ccm/openjdkhead/test/jdk/java/lang/invoke/lambda/TestLambdaFileEncodingSerialization.java:38: error: error while writing ABC??: bad filename RelativeFile[TestLambdaFileEncodingSerialization$ABC??.class]
	public static class ABC?? implements Serializable {
	              ^
Note: /home/admin/mydocker/ccm/openjdkhead/test/jdk/java/lang/invoke/lambda/TestLambdaFileEncodingSerialization.java uses unchecked or unsafe operations.
Note: Recompile with -Xlint:unchecked for details.
1 error
```
To avoid this (possible javac bug!), is there a different way to re-create this problem with the classname containg UTF chars?

pshipton:
Definitely a javac bug. 

While we could craft something this way that works for you, it won't work across all systems having different default character sets. The bytes I used could be many other non-ascii values, a single non-ascii byte would be fine. The non-ascii byte(s) used need to be valid to the file system, and also translate to a unicode character which is valid in a class name. What is the character set of the system you are using? 

To avoid file system problems and make something that always works, you would generate the lambda class source and call javac programmatically if that is possible, with a byte[] or String (converted from byte[] not using UTF8). Getting back class bytes, the class can be created programmatically without using the file system at all.

pshipton:
I suggested this in Slack already, trying setting `-J-Dsun.jnu.encoding=UTF-8` assuming the machines is UTF-8 (environment variable LANG=en_US.UTF-8 or similar) as sun.jnu.encoding controls the file name encoding.

andrew-m-leonard:
Fix has been contributed upstream to openjdk: https://hg.openjdk.java.net/jdk/jdk/rev/335ae2826fbf

andrew-m-leonard:
@DanHeidinga @pshipton @mikezhang1234567890 fyi, contributed to openjdk

pshipton:
jenkins test sanity zlinux jdk11

PR #307, opened 2020-05-18
Openj9 only javadoc

lumpfish:

PR #320, opened 2020-06-23
Create javadoc for only openj9 extensions

lumpfish:
Signed-off-by: Simon Rushton <rushton@uk.ibm.com>

### DIFF
--- a/closed/custom/Docs.gmk
+++ b/closed/custom/Docs.gmk
@@ -25,11 +25,10 @@ JVMTI_HTML := $(TOPDIR)/closed/variant-server/gensrc/jvmtifiles/jvmti.html
 
 OPENJ9_BASE_URL := https://www.eclipse.org/openj9/
 OPENJ9_BUG_SUBMIT_URL := https://github.com/eclipse/openj9/issues
-OPENJ9_FULL_COMPANY_NAME := IBM Corp
-OPENJ9_JAVADOC_BOTTOM := \
-    <a href="$(OPENJ9_BASE_URL)">Link</a> to Eclipse OpenJ9 documentation.<br> \
-    To raise a bug report or suggest an improvement create an <a href="$(OPENJ9_BUG_SUBMIT_URL)">Eclipse Openj9 issue.</a><br> \
-    Copyright &copy; 1998, $(COPYRIGHT_YEAR), $(OPENJ9_FULL_COMPANY_NAME). All rights reserved.
+OPENJ9_JAVADOC_BOTTOM = \
+    <a href="$(OPENJ9_BASE_URL)" target="_blank">Eclipse OpenJ9 website.</a><br> \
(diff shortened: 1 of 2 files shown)


lumpfish:
This PR follows on from the backed out #307 which did not build on Windows.

closed/custom/Docs.gmk
```diff
@@ -112,7 +110,199 @@ $(OPENJ9_JAVADOC_MARKER) : $(OPENJDK_JAVADOC_MARKER)
...
+# file which is created during the full jdk javadoc generation exists before running
+# javadoc for just the openj9 files.
+
+# If running on Windows, The source file names in the response file must be converted to
+# Windows format before passing them to javadoc.
+OPENJ9_ONLY_JAVADOC_FIXPATH_FILELIST := $(foreach file,$(OPENJ9_ONLY_JAVADOC_FILELIST),$(call FixPath,$(file)))
+
+$(OPENJ9_ONLY_JAVADOC_MARKER) : $(OPENJDK_JAVADOC_MARKER) $(OPENJ9_ONLY_FILELIST)
+	$(call LogWarn, Generating javadoc for openj9 source files)
+	$(ECHO) '$(OPENJ9_ONLY_JAVADOC_SCRIPT_BODY_ESC)' >$(OPENJ9_ONLY_JAVADOC_SCRIPT)
```
keithc-ca, line 295: In the previous version, this line was immediately before the use by bash (now line 306). I don't think it's wrong here - just curious why it moved?

lumpfish: Ah yes.
I moved it there so that the file OPENJ9_ONLY_JAVADOC_FILES would be written to the script file with its cygwin file name because I was then converting that file name to Windows format before passing it to javadoc on the javadoc command line. After adding some logging and examining the output I realised that windows command lines are actually run via fixpath.exe which (I now assume) converts all cygwin paths in the command line to windows format before executing the command.  So the the file name conversion I had done was unnecessary - and I then forgot to put the line writing out the script code back to where it was originally.

keithc-ca: Thanks for the explanation. I'll let you decide whether you want to move it back or not.

lumpfish: I prefer it in its original place, immediately before the line which executes the script.  I've updated the PR.

keithc-ca:
Jenkins compile aix,osx,plinux,xlinux,zlinux,win jdk11

keithc-ca:
The AIX build failed due to eclipse/openj9#9997; the plinux failure looks like an infrastructure problem.

keithc-ca:
Jenkins compile aix,plinux jdk11

keithc-ca:
The plinux build failed due to a problem uploading results to artifactory - the build itself succeeded.
@lumpfish Please squash in preparation for merging.

PR #323, opened 2020-06-29
Enable cmake by default on p and zlinux

dnakamura:

### DIFF
--- a/closed/autoconf/custom-hook.m4
+++ b/closed/autoconf/custom-hook.m4
@@ -60,7 +60,7 @@ AC_DEFUN([OPENJ9_CONFIGURE_CMAKE],
     ],
     [
       case "$OPENJ9_PLATFORM_CODE" in
-        oa64|xa64)
+        oa64|xa64|xl64|xz64)
           with_cmake=cmake
           ;;
         *)


closed/autoconf/custom-hook.m4
```diff
@@ -60,7 +60,7 @@ AC_DEFUN([OPENJ9_CONFIGURE_CMAKE],
     ],
     [
       case "$OPENJ9_PLATFORM_CODE" in
-        oa64|xa64)
```
pshipton, line 63: This doesn't affect the XL builds, is that intensional?

keithc-ca: `OPENJ9_PLATFORM_CODE` is independent of normal/XL.

pshipton: When I look at the nightly builds for xlinux, it seems the normal build is using cmake and the XL build is not.

dnakamura: are you sure you are looking at jdk11? Looks like cmake to me https://ci.eclipse.org/openj9/job/Build_JDK11_x86-64_linux_xl_Nightly/423/console

pshipton: Yes, my mistake. I was looking at jdk11 but at a plinux instead of xlinux build.

PR #327, opened 2020-07-14
Expand content of debug images

keithc-ca:
Also include OpenJ9 `.map` files on Windows; see eclipse/openj9#10099.


keithc-ca:
Jenkins test sanity aix,osx,plinux,xlinux,zlinux,win jdk11

PR #332, opened 2020-07-28
Don't default to CMake on cross compiles

dnakamura:
CMake builds don't currently support cross compiles


pshipton:
jenkins compile zlinux jdk11

pshipton:
@knn-k fyi

PR #340, opened 2020-08-25
Include openj9.cuda and openj9.gpu in JRE

keithc-ca:
Issue: eclipse/openj9#10030.


pshipton:
PR testing doesn't build the JRE, but this has been tested manually.

PR #347, opened 2020-09-23
Enable cmake by default on aix

dnakamura:


dnakamura:
seeing test failure in `TestAttachErrorHandling_0`, could be eclipse/openj9#3081

dnakamura:
it does appear to be the same issue, and occur at roughly the same frequency as the uma builds (based on a couple of grinder runs).

pshipton:
jenkins test sanity,extended,sanity.system,extended.system,sanity.openjdk aixcm jdk11

keithc-ca (approved):

PR #353, opened 2020-10-14
Remove references to 'com.ibm.oti.vm.library.version'

keithc-ca:
* don't define unused preprocessor macro `com.ibm.oti.vm.library.version`
* remove unnecessary read permission for `com.ibm.oti.vm.library.version`

Depends on eclipse/openj9#10888.

### DIFF
--- a/src/java.base/share/lib/security/default.policy
+++ b/src/java.base/share/lib/security/default.policy
@@ -226,7 +226,6 @@ grant codeBase "jrt:/openj9.cuda" {
     permission java.lang.RuntimePermission "loadLibrary.cuda4j29";
     permission java.lang.RuntimePermission "modifyThread";
     permission java.lang.RuntimePermission "modifyThreadGroup";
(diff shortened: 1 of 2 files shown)


src/java.base/share/lib/security/default.policy
```diff
@@ -226,7 +226,6 @@ grant codeBase "jrt:/openj9.cuda" {
     permission java.lang.RuntimePermission "loadLibrary.cuda4j29";
     permission java.lang.RuntimePermission "modifyThread";
     permission java.lang.RuntimePermission "modifyThreadGroup";
-    permission java.util.PropertyPermission "com.ibm.oti.vm.library.version", "read";
```
pshipton, line 229: I don't think this should be removed. Access has always been allowed to this property, removing the access would break customers who are already using it.

pshipton: Nm, I should have looked at the context.

pshipton (approved):

PR #355, opened 2020-10-22
Improve integration with openjdk build system

keithc-ca:
The primary change here is to respect the `JOBS` parameter (on the `make` command line or via an explicit or implicit configuration choice for `--with-jobs=n`). Prior to this change, a build would run up to `2n-1` jobs concurrently (`n-1` for openjdk and `n` for openj9). With this, cmake builds as a whole will operate within that limit. Fixing UMA builds similarly would be significantly more work and, given our plans to stop using UMA, I didn't consider that effort worthwhile.

* Interact properly with jobserver by using '+' prefix on recursive make commands.
* `MAKEFLAGS` inherited from openjdk includes options (e.g. '-r' and '-R') that don't work well with OpenJ9 and OMR when not using cmake. When using cmake, however, retaining the value inherited from openjdk should improve build performance (e.g. due to presence of '-r' and '-R' which avoids the costs due to built-in rules and macros).
* Log commands started by cmake when appropriate.
* Use `$(MAKE_ARGS)` for consistency with openjdk makefiles.
* Delay expansion of `$(OUTPUTDIR)`.
* Fold `build-openj9-tools` into `generate-j9jcl-sources`.
* Add required dependency on `generate-j9jcl-sources`.
* Add `EXTRA_CMAKE_ARGS` last to enable it to override other cmake options.
* Use `ECHO` instead of `PRINTF` when possible.
* Use `OPENJDK_TARGET_OS` instead of `HOTSPOT_TARGET_OS`.
* Stop exporting `CMAKE`, `OPENJ9_ENABLE_CMAKE` to avoid warning for 'make reconfigure':
```
    WARNING: Ignoring value of CMAKE from the environment.
```

### DIFF
--- a/closed/OpenJ9.gmk
+++ b/closed/OpenJ9.gmk
@@ -54,8 +54,21 @@ ifeq (,$(OPENJ9OMR_SHA))
   $(error Could not determine OMR SHA)
 endif
 
-# openjdk makeflags don't work with openj9/omr native compiles; override with number of CPUs which openj9 and omr need supplied
-override MAKEFLAGS := -j $(JOBS)
+ifeq (true,$(OPENJ9_ENABLE_CMAKE))
+  # If a logging level was specified that shows commands, tell cmake to do likewise.
+  ifneq (,$(or $(filter true,$(LOG_CMDLINES)),$(filter debug trace,$(LOG_LEVEL))))
+    MAKE_ARGS += VERBOSE=ON
+  endif
+else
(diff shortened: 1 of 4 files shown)


pshipton:
jenkins test sanity,extended aix,plinux,osx.win jdk11

keithc-ca:
Jenkins test sanity,extended aix,plinux,osx,win jdk11

pshipton (approved):

PR #356, opened 2020-10-28
CMake: Pass jdk version to cmake

dnakamura:


dnakamura:
see ibmruntimes/openj9-openjdk-jdk#243

pshipton (approved):

PR #358, opened 2020-11-05
Add vm.bits property

sophia-guo:
Signed-off-by: Sophia Guo <sophia.gwf@gmail.com>

### DIFF
--- a/closed/test/jtreg-ext/requires/OpenJ9PropsExt.java
+++ b/closed/test/jtreg-ext/requires/OpenJ9PropsExt.java
@@ -1,6 +1,6 @@
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2019, 2019 All Rights Reserved
+ * (c) Copyright IBM Corp. 2019, 2020 All Rights Reserved
  * ===========================================================================
  * 
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


pshipton:
jenkins test sanity.openjdk all jdk11

pshipton:
jenkins test sanity.openjdk plinux,plinuxxl,xlinux,xlinuxxl jdk11

closed/test/jtreg-ext/requires/OpenJ9PropsExt.java
```diff
@@ -31,14 +31,30 @@
...
+    }
+    /**
+     * @return VM bitness, the value of the "sun.arch.data.model" property.
+     */
+    protected String vmBits() {
+        String dataModel = System.getProperty("sun.arch.data.model");
+        if (dataModel != null) {
+            return dataModel;
+        } else {
+            return errorWithMessage("Can't get 'sun.arch.data.model' property");
```
pshipton, line 56: Wouldn't it be better to throw an exception and have the test harness fail, than to set the test property to an error message that will just cause the test not to run?

sophia-guo: The class could set any  properties used by @requires expression. If the vm property setting error fails the test harness that means none of any tests will be run. Setting to an error message will not block all other tests. It will skip  the tests with that specific property @requires expression.

I guess it also depends on how important that property string is. If the property 'sun.arch.data.model' is null should we suppose that JDK is broken with an unacceptable failure? If no, we should not fail the test harness due to that. If it's yes I think we should cover this scenario by tests.  And looks like it is covered by test/hotspot/jtreg/gc/TestObjectAlignment.java.

pshipton: My argument is that it is better to fail and run none of the tests, which indicates there is something wrong that needs to be fixed. Otherwise only certain tests fail to run, which may or may not be noticed for some time.

> If the property 'sun.arch.data.model' is null should we suppose that JDK is broken with an unacceptable failure

Seems reasonable to me. It is and should always be set. Since `test/hotspot/jtreg/gc/TestObjectAlignment.java` is a hotspot test, I don't expect it's run on OpenJ9.

pshipton: Also, there is no obvious connection between TestObjectAlignment.java failing and openjdk tests not running.

pshipton: @sophia-guo pls see my previous comments.

sophia-guo: @pshipton yes, will update. Sorry, I missed the comments.

sophia-guo:
@pshipton Updated. Thanks.

pshipton (approved):

pshipton:
jenkins test sanity.openjdk plinux,aix,osx,win jdk11

pshipton:
@sophia-guo pls update jdk15 and jdk (next) as well.

sophia-guo:
Yes, will do.

PR #363, opened 2020-11-17
Copy jni_convert.h from source tree

dnakamura:
Copy from source tree rather than build tree as the file does not exist
in the build tree for cmake builds.


keithc-ca (approved):

keithc-ca:
This only affects z/OS: no value in running PR testing here.

PR #369, opened 2020-12-07
Do not rely on usingSharedClasses() to add url to loaderURLs

hangshao0:
When reaching URLClassPath.getLoader() before module system is
initialized, usingSharedClasses() always returns false. It is
likely that SCC support is not initialized for BuiltinClassLoaders
yet, which is done during module system initialization. We still need
to update loaderURLs in this case.

Fixes https://github.com/eclipse/openj9/issues/11342

Signed-off-by: Hang Shao <hangshao@ca.ibm.com>

### DIFF
--- a/src/java.base/share/classes/jdk/internal/loader/URLClassPath.java
+++ b/src/java.base/share/classes/jdk/internal/loader/URLClassPath.java
@@ -24,7 +24,7 @@
  */
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 1997, 2019 All Rights Reserved
+ * (c) Copyright IBM Corp. 1997, 2020 All Rights Reserved
  * ===========================================================================
  */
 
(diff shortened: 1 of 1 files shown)


hangshao0:
Manfully tested, the `IllegalStateException` goes away with this change.

src/java.base/share/classes/jdk/internal/loader/URLClassPath.java
```diff
@@ -569,11 +570,27 @@ private synchronized Loader getLoader(int index) {
...
+                if (!VM.isModuleSystemInited()) {                                        //OpenJ9-shared_classes_misc
+                    /**                                                                  //OpenJ9-shared_classes_misc
+                     * usingSharedClasses() is false. If Module System is not            //OpenJ9-shared_classes_misc
+                     * initialized yet, it is likely that BuiltinClassLoader             //OpenJ9-shared_classes_misc
+                     * hasn't called setSharedClassProvider(). We still nedd to          //OpenJ9-shared_classes_misc
+                     * update loaderURLs in this case.                                   //OpenJ9-shared_classes_misc
+                     */                                                                  //OpenJ9-shared_classes_misc
+                    if (null != loaderURLs) {                                            //OpenJ9-shared_classes_misc
+                        /* update search path URLs */                                    //OpenJ9-shared_classes_misc
+                        loaderURLs.add(url);                                             //OpenJ9-shared_classes_misc
```
pshipton, line 587: This is common code, the following can avoid duplicating it.
```
if (null == loaderURLs) { 
    loaderURLs = new ArrayList(path.size());
}
loaderURLs.add(url);
```

hangshao0: Fixed.

pshipton:
jenkins test sanity zlinux jdk11

pshipton (approved):

hangshao0:
> Will create PR for jdk15 and jdkNext repos once this is reviewed and merged.

https://github.com/ibmruntimes/openj9-openjdk-jdk15/pull/49
https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/256

PR #260, opened 2020-01-31
Fix ludcl refresh logic

theresa-m:
Fixes: https://github.com/eclipse/openj9/issues/8454

Signed-off-by: Theresa Mammarella <Theresa.T.Mammarella@ibm.com>

PR #370, opened 2020-12-10
WIP: Add nested resolveClass checks and ludcl refresh logic

sharon-wang:
Fixes: https://github.com/eclipse/openj9/issues/9912, where a `ClassNotFoundException` was occurring due to a missing refresh of the cached last user-defined class loader (ludcl) in a case where nested deserialization was involved.

This PR adds extra checks in `resolveClass()` to refresh the ludcl in a nested situation, following the existing implementation in `readObject()` that deals with nested calls.

https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/5e81062d6d3f850c4167ffc705e5cab3f273dd50/src/java.base/share/classes/java/io/ObjectInputStream.java#L542-L602

### DIFF
--- a/src/java.base/share/classes/java/io/ObjectInputStream.java
+++ b/src/java.base/share/classes/java/io/ObjectInputStream.java
@@ -550,9 +550,9 @@ public class ObjectInputStream
             throw new AssertionError("internal error");
 
         ClassLoader oldCachedLudcl = null;
-	boolean setCached = false;
+        boolean setCached = false;
 
-	if (((null == curContext) || refreshLudcl) && (isClassCachingEnabled)) {
+        if (((null == curContext) || refreshLudcl) && (isClassCachingEnabled)) {
             oldCachedLudcl = cachedLudcl;
 
             // If caller is not provided, follow the standard path to get the cachedLudcl.
@@ -866,13 +866,22 @@ public class ObjectInputStream
         throws IOException, ClassNotFoundException
     {
         String name = desc.getName();
+        ClassLoader oldCachedLudcl = null;
+        boolean setCached = false;
+
         try {
             if (null == classCache) {
                 return Class.forName(name, false, latestUserDefinedLoader());
             } else {
                 if (refreshLudcl) {
+                    oldCachedLudcl = cachedLudcl;
                     cachedLudcl = latestUserDefinedLoader();
+
+                    setCached = true;
                     refreshLudcl = false;
+                    if (null == startingLudclObject) {
+                        startingLudclObject = this;
+                    }
                 }
                 return classCache.get(name, cachedLudcl);
             }
@@ -882,6 +891,15 @@ public class ObjectInputStream
                 return cl;
             } else {
                 throw ex;
+            }
+        } finally {
+            /* Back to the start, refresh ludcl cache on next call. */
+            if (this == startingLudclObject) {
+                refreshLudcl = true;
+                startingLudclObject = null;
+            }
+            if (setCached) {
+                cachedLudcl = oldCachedLudcl;
             }
         }
     }


tajila:
Is the failure J9 specific or did it occur on hotspot as well?

sharon-wang:
It's J9-specific. It was due to an issue around `-Dcom.ibm.enableClassCaching`.

sharon-wang:
FYI @vijaysun-omr -- do we want to do some perf testing for this change? Perhaps DayTrader and throughput benchmarks similar to https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/341.

tajila:
@pshipton can you please review these changes

vijaysun-omr:
Hi @sharon-wang , I agree with the testing suggestion that you had. Please ask Piyush if he can help with doing the Daytrader7 runs.

pshipton:
jenkins test sanity,extended,sanity.system,extended.system.sanity.openjdk zlinux jdk11

pshipton:
Changes lgtm. All the testing passed. I won't merge until we get the perf results. We'll need the same changes for jdk15 (if it's going to make the final release), jdk16 (a new repo should be created shortly) and jdk (next).

sharon-wang:
LibertyStartupDT7 Results
---
These are the aggregate startup results from 6 iterations of the benchmark.
|   | Mean Startup Time (ms) | Median Startup Time (ms) | Confidence Interval |
|--:|:-:|:-:|:-:|
| **with new changes** | 3522.9 | 3509.0 | 0.0055 |
| **no changes** | 3477.0 | 3469.0 | 0.0061 |
|  **diff** | +45.9 | +40 |
|  **diff %** | +1.32% | +1.15% |

---

I am running a throughput benchmark separately and expect to have results in the next couple of days.

sharon-wang:
Throughput Benchmark Results
---
4 iterations of the benchmark were run.
|   | max_jOPS Mean | critical_jOPS Mean |
|--:|:-:|:-:|
|  **diff %** | -3.64% | -4.44% |

This seems like a somewhat sizeable regression. @vijaysun-omr I can send you the full results via message for more details, if needed.

vijaysun-omr:
I agree @sharon-wang. This needs further investigation but perhaps also verification by Piyush and maybe more importantly by Gita Koblents and Julian Wang since SPECjbb2015 is most relevant on the Power platform where publications are done/planned.

sharon-wang:
Thanks Vijay, I'll reach out to them for further investigation. The above throughput results were from a Linux AMD machine, so I'll kick off another run on PPC.

vijaysun-omr:
Sharon, please also be aware that the runs on Power that I mentioned ought to also happen with the command line options that we publish SPEC results with on that platform. Of course you can do a sniff test on Power yourself using standard options but you will probably need the Power team's help to get on a setup that is relevant from an external standpoint.

sharon-wang:
Ah I see, I'll coordinate with the Power team so we can get the relevant setup to measure throughput.

sharon-wang:
Throughput Benchmark Results (_Linux Power_)
---
10 iterations of the benchmark were run.
|   | max_jOPS Mean | critical_jOPS Mean |
|--:|:-:|:-:|
|  **diff %** | -2.22% | -3.29% |

OpenJ9: https://github.com/eclipse/openj9/commit/19fefcf952d
OMR: https://github.com/eclipse/openj9-omr/commit/4c466e6429c
Baseline JDK 11: 6319cd4ec97b55a1943b79598ffaf49ebf074338
PR Changes JDK 11: Baseline + 286e3cccbdad5cbddfa17fd8fbe59a211cf79ac2 from this PR.

Although a bit less of a regression than the results from https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/370#issuecomment-746631582, there still appears to be a noticeable regression in throughput.

zl-wang:
do you have the throughput number when ludcl caching is disabled? just to get a current perspective ...

i remembered the performance gap was huge without caching for 2 reasons:  1) ludcl needs stackWalking;  2) class.forName had locking (this one might have been fixed).

so, for an ideal fix, refreshing cache should be shepherded to the scenarios where it is really needed.  i.e.  for jbb2015, there is no nested classes involved (did i read it right?), there should be no impact.

sharon-wang:
I can do another set of runs with ludcl caching disabled. Since ludcl caching is enabled by default, I hadn't tested with it off in the above results.

sharon-wang:
> I can do another set of runs with class caching disabled (`com.ibm.enableClassCaching=false`). Since ludcl caching is enabled by default with class caching, I hadn't tested with class caching off in the above results.

Hmm, doing a set of runs with class caching disabled altogether is likely going to give us significantly reduced throughput, I assume. Not sure if those results will illustrate the difference we're looking for, since it disables more than ludcl caching.

vijaysun-omr:
@zl-wang any further guidance for Sharon ?

zl-wang:
it is a question needing answer whether the bug-inducing scenario happens in jbb2015, i.e. whether the cached content is still valid, and the current fix threw it away unnecessarily. then, we can consider how to tighten the fix aiming to exactly match the buggy scenario.

the previous ask for ludcl caching disabled throughput was to get a sense where the current fix stood re throwing away the cache.

sharon-wang:
> it is a question needing answer whether the bug-inducing scenario happens in jbb2015, i.e. whether the cached content is still valid, and the current fix threw it away unnecessarily. then, we can consider how to tighten the fix aiming to exactly match the buggy scenario.

I don't think that the bug-inducing scenario happens in jbb2015. I was unable to pinpoint the exact cause of the bug as the code is proprietary. When trying different fixes, refreshing the ludcl on every call to `resolveClass()` was working, and refreshing only for nested `resolveClass()` calls was a bit of a tighter fix, although it is doing some extra refreshing as we see in the throughput results.

It seems that jbb2015 has some amount of nesting occurring which is hitting the fix, but not the same nesting situation that would lead to the bug. As such, there is some unnecessary refreshing of the cached ludcl, which is leading to the reduction in throughput.

Previously, there was an added fix for nested calls to `readObject()` (https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/260). Perhaps there's some better coordination between changes in this PR and the `readObject()` fix that can make the new fix tighter -- looking into this.

sharon-wang:
I'm working on some alternatives and will be working with the user to confirm functionality is working. Once I have some working builds, I'll run the throughput benchmarks again to check if the perf gap is being closed, and will update this PR.

dnakamura:
I'm currently looking at optimizations limiting the ammount of stack we need to walk, since in a recursive situation we only really need to walk back as far as  any previous calls where we already checked for ludcl.  Currently  we do have functionality in the vm for a truncated stack walk as part of the ORB code. I'm currently experimenting with it to see if it can be reused here without breaking anything.

PR #374, opened 2021-01-07
Update use of buildj9tools.mk to build java preprocessor

keithc-ca:
For consistency with newer Java versions.
Depends on eclipse/openj9#11589.


keithc-ca:
Jenkins test sanity osx,win jdk11

pshipton (approved):

PR #381, opened 2021-01-29
Fix DDR-related build issues

keithc-ca:
* avoid race creating directories when DDR is enabled
* do nothing for openj9.dtfj-gensrc-src when DDR is disabled

See: eclipse/openj9#11818; depends on eclipse/openj9#11824.


pshipton:
jenkins compile zlinux jdk11

pshipton (approved):

PR #382, opened 2021-02-04
Default jdk11 to build mixedrefs, unless cross compiling 

pshipton:
Prepare for switching jdk11 to build mixedrefs by default. This is dependent on https://github.com/AdoptOpenJDK/TKG/pull/155 being merged first.

### DIFF
--- a/closed/autoconf/custom-hook.m4
+++ b/closed/autoconf/custom-hook.m4
@@ -1,5 +1,5 @@
 # ===========================================================================
-# (c) Copyright IBM Corp. 2017, 2020 All Rights Reserved
+# (c) Copyright IBM Corp. 2017, 2021 All Rights Reserved
 # ===========================================================================
 # This code is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License version 2 only, as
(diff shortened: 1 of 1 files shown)


pshipton:
@llxia fyi

closed/autoconf/custom-hook.m4
```diff
@@ -358,23 +358,24 @@ AC_DEFUN([OPENJ9_PLATFORM_SETUP],
...
+      OPENJ9_LIBS_SUBDIR=default
     else
+      OPENJ9_BUILD_MODE_ARCH="${OPENJ9_CPU}_cmprssptrs"
+      OPENJ9_LIBS_SUBDIR=compressedrefs
+    fi
+  else
+    OMR_MIXED_REFERENCES_MODE=static
+    if test "x$with_mixedrefs" = xdynamic; then
+      OMR_MIXED_REFERENCES_MODE=dynamic
+    elif test "x$with_mixedrefs" != xyes -o "x$with_mixedrefs" != xstatic; then
```
keithc-ca, line 374: * the extra `-o` on line 361 is a problem
* nit: please use a space on both sides if `;` in `if` statements

I think it is easier to follow with fewer negative tests:
```
  if test "x$with_noncompressedrefs" = xyes -o "x$with_mixedrefs" = xno ; then
    OMR_MIXED_REFERENCES_MODE=off
    if test "x$with_noncompressedrefs" = xyes ; then
      OPENJ9_BUILD_MODE_ARCH="${OPENJ9_CPU}"
      OPENJ9_LIBS_SUBDIR=default
    else
      OPENJ9_BUILD_MODE_ARCH="${OPENJ9_CPU}_cmprssptrs"
      OPENJ9_LIBS_SUBDIR=compressedrefs
    fi
  else
    if test "x$with_mixedrefs" = xstatic - o "x$with_mixedrefs" = xyes ; then
      OMR_MIXED_REFERENCES_MODE=static
    elif test "x$with_mixedrefs" = xdynamic ; then
      OMR_MIXED_REFERENCES_MODE=dynamic
    else
      AC_MSG_ERROR([OpenJ9 supports --with-mixedrefs=static and --with-mixedrefs=dynamic])
    fi
    OPENJ9_BUILD_MODE_ARCH="${OPENJ9_CPU}_mxdptrs"
    OPENJ9_LIBS_SUBDIR=default
  fi
```

pshipton: Makes sense. The with_mixedrefs part is basically the same as the original, not sure why I felt the need to change it.

pshipton: Ah, because it doesn't work. Another update coming.

pshipton: I've pushed a change that makes it correct although it's a bit funny. I'll let you comment on it.

closed/autoconf/custom-hook.m4
```diff
@@ -358,23 +358,26 @@ AC_DEFUN([OPENJ9_PLATFORM_SETUP],
...
+    if test "x$with_noncompressedrefs" = xyes ; then
+      OPENJ9_BUILD_MODE_ARCH="${OPENJ9_CPU}"
+      OPENJ9_LIBS_SUBDIR=default
+    else
+      OPENJ9_BUILD_MODE_ARCH="${OPENJ9_CPU}_cmprssptrs"
+      OPENJ9_LIBS_SUBDIR=compressedrefs
+    fi
+  else
+    OMR_MIXED_REFERENCES_MODE=static
+    if test "x$with_mixedrefs" = xyes -o "x$with_mixedrefs" = xstatic ; then
```
keithc-ca, line 372: This needs another clause when `--with-mixedrefs` is not specified:
```
    if test "x$with_mixedrefs" = x -o "x$with_mixedrefs" = xyes -o "x$with_mixedrefs" = xstatic ; then
```

closed/autoconf/custom-hook.m4
pshipton: Actually it still doesn't work. It needs to handle the case where --with_mixedrefs isn't set.

pshipton: Hopefully the latest update is good.

closed/autoconf/custom-hook.m4
pshipton: Just pushed that.

keithc-ca (approved):

pshipton:
I'll just make sure it actually builds.

jenkins compile zlinux jdk11

keithc-ca:
With this new default for mixedrefs, using `--with-cmake=no` will now require the additional option `--with-mixedrefs=no`.
@pshipton I hope you'll agree that's not too onerous.

pshipton:
> I hope you'll agree that's not too onerous.

Seems fine considering this stuff is temporary, in that once we're all switched over and convinced everything is working we no longer need to support non-cmake and non-mixed builds.

pshipton:
I don't know that the project uses `--with-cmake=no` anywhere. We could update some doc, but not sure there is any.

pshipton:
Is there a way to tell from the build output that mixedrefs is building? I checked the build itself and there is no compresedrefs dir even though the -version output shows compressed, so I'm convinced it's working.

keithc-ca:
`-with-cmake=no` is equivalent to `--without-cmake` which is used in [default.yml](https://github.com/eclipse/openj9/blob/master/buildenv/jenkins/variables/defaults.yml#L142) which is augmented with `--with-mixedrefs=no` by eclipse/openj9#11882.

keithc-ca:
`-version` output speaks to _whether_ compressed references are being used, not _how_ the build is packaged.

keithc-ca:
The presence of the `_full` shared library variants signals a mixed references build.
```
jdk/lib/default/libj9gc29.so
jdk/lib/default/libj9gc_full29.so
```

pshipton:
jenkins compile win jdk11

pshipton:
@keithc-ca let me know if you are happy with this change and I'll update the other versions as well.

keithc-ca:
Jenkins copyright check

keithc-ca:
> @keithc-ca let me know if you are happy with this change and I'll update the other versions as well.

The latest test looks good to me.

ThyThy2505 (commented):
keithc-ca:uma_no_mixedrefs

PR #385, opened 2021-02-11
Access nonce only when non-null

keithc-ca:
This is a back-port of ibmruntimes/openj9-openjdk-jdk#288 for jdk11.


keithc-ca:
Updated to match ibmruntimes/openj9-openjdk-jdk#288.

pshipton (approved):

PR #389, opened 2021-03-12
Improve backward compatibility of openj9.dtfj

keithc-ca:
This is part of the solution to eclipse/openj9#12182; it depends on eclipse/openj9#12210.

* compile using only those constants guaranteed to be available
* include `CompatibilityConstants29.dat` in `openj9.dtfj` module


pshipton:
jenkins compile all jdk11 depends eclipse/openj9#12210

pshipton:
jenkins compile all jdk11 depends eclipse/openj9#12210

pshipton (approved):

PR #392, opened 2021-03-17
Patch the DDR tools if present in the boot jdk

keithc-ca:
See eclipse/openj9#12229.


pshipton:
jenkins compile zlinux jdk11

PR #397, opened 2021-03-31
Remove redundant launcher specifications of MAIN_MODULE

keithc-ca:
`MAIN_MODULE` only needs to be specified when the main class is not in the current module.

See https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/make/launcher/LauncherCommon.gmk#L82.


keithc-ca:
The current module is implicit in the name of the file (`Launcher-{module}.gmk`). Most launchers use classes from the same module, the exception is `pack200` - see https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/make/launcher/Launcher-jdk.pack.gmk#L29.

pshipton (approved):

PR #404, opened 2021-05-05
Handle Exceptional Padding Cases in the Native RSADP Verification

samasri:
This PR fixes part of eclipse-openj9/openj9#11500 by changing the way verification is implemented in the RSADP operation.

More specifically, this PR adds a special case when verifying a message. For certain key sizes (where keySize % 8 == 1), the decrypted message retrieved from the `RSA_public_decrypt` includes a 1 byte padding at the beginning of the message. In these cases, this padding must be ignored when comparing the decrypted message to the original message.

Signed-off-by: Samer AL Masri <samasri@ibm.com>

### DIFF
--- a/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
+++ b/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
@@ -1,6 +1,6 @@
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2020 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2021 All Rights Reserved
  * ===========================================================================
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -1661,7 +1661,7 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSADP
     msg_len = (*OSSL_RSA_private_encrypt)(kLen, kNative, mNative, rsaKey, RSA_NO_PADDING);
 
     if ((-1 != verify) && (-1 != msg_len)) {
-        if (verify == kLen) {
+        if ((verify == kLen) || (verify == (kLen + 1))) {
             k2 = malloc(kLen * (sizeof(unsigned char)));
             if (NULL != k2) {
 
@@ -1670,10 +1670,29 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSADP
                 if (-1 != msg_len2) {
 
                     int i;
-                    for (i = 0; i < verify; i++) {
-                        if (kNative[i] != k2[i]) {
+                    /*
+                     * For certain key sizes, the decrypted message retrieved from the RSA_public_decrypt
+                     * includes a 1 byte padding at the beginning of the message. In these cases, this
+                     * padding must be zero. And the comparison to the original message should not include
+                     * this first byte.
+                     */
+                    if (verify == (kLen + 1)) {
+                        if (0 != k2[0]) {
                             msg_len = -2;
-                            break;
+                        } else {
+                            for (i = 0; i < kLen; i++) {
+                                if (kNative[i] != k2[i + 1]) {
+                                    msg_len = -2;
+                                    break;
+                                }
+                            }
+                        }
+                    } else { // if verify == kLen
+                        for (i = 0; i < verify; i++) {
+                            if (kNative[i] != k2[i]) {
+                                msg_len = -2;
+                                break;
+                            }
                         }
                     }
                 } else {


samasri:
@pshipton Could you please review this?

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
```diff
@@ -138,17 +140,40 @@ public static int getByteLength(BigInteger b) {
...
+        RSADPBigNumRPointer() {
+            this.rsadpBN = nativeCrypto.create_rsadp_BN();
+        }
+
+        @Override
+        protected synchronized void finalize() throws Throwable {
+            try {
+                nativeCrypto.free_rsadp_BN(rsadpBN);
+            } finally {
+                super.finalize();
```
pshipton, line 171: This is unnecessary, it doesn't do anything.

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
```diff
@@ -138,17 +140,40 @@ public static int getByteLength(BigInteger b) {
...
+
+    static class RSADPBigNumRPointer {
+        long rsadpBN = 0;
+
+        RSADPBigNumRPointer() {
+            this.rsadpBN = nativeCrypto.create_rsadp_BN();
+        }
+
+        @Override
+        protected synchronized void finalize() throws Throwable {
```
pshipton, line 167: What is this synchronized with? Nothing else is synchronized.

pshipton:
jenkins compile win jdk11

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -47,6 +48,46 @@
 # include <pthread.h>
 #endif /* defined(WINDOWS) */
 
+// This definition is needed to compile the "OSSL_RSA_get0_key_compatibility" function
+// Since the first parameter of that function is of type RSA, which is
+// a typedef of rsa_st, the function would not compile without defining
+// rsa_st here
+struct rsa_st {
```
keithc-ca, line 55: This is not a good idea; `rsa_st` is intended to be an opaque type - see https://github.com/openssl/openssl/issues/1604#issuecomment-248237417.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -347,10 +413,21 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
...
+    OSSL_BN_mod_exp = (OSSL_BN_mod_exp_t *)find_crypto_symbol(crypto_library, "BN_mod_exp");
+    OSSL_BN_ucmp = (OSSL_BN_ucmp_t *)find_crypto_symbol(crypto_library, "BN_ucmp");
+    OSSL_BN_clear = (OSSL_BN_zero_t *)find_crypto_symbol(crypto_library, "BN_clear");
+    OSSL_BN_CTX_new = (OSSL_BN_CTX_new_t *)find_crypto_symbol(crypto_library, "BN_CTX_new");
+    OSSL_BN_CTX_free = (OSSL_BN_CTX_free_t *)find_crypto_symbol(crypto_library, "BN_CTX_free");
+    OSSL_BN_new = (OSSL_BN_new_t *)find_crypto_symbol(crypto_library, "BN_new");
     OSSL_BN_set_negative = (OSSL_BN_set_negative_t *)find_crypto_symbol(crypto_library, "BN_set_negative");
     OSSL_BN_free = (OSSL_BN_free_t *)find_crypto_symbol(crypto_library, "BN_free");
 
+    if(OSSL_RSA_get0_key == NULL) {
```
keithc-ca, line 427: formatting: space after `if`

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1626,6 +1710,55 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSAEP
...
+/* RSADP Cryptographic Primitive, RSA Private Key operation
+ * Returns -1 on error
+ * The param verify is -1 for 'no verify', otherwise it is size of m (with verify)
+ *
+ * Class:     jdk_crypto_jniprovider_NativeCrypto
+ * Method:    RSADP
+ * Signature: ([BI[BIJ)I
+ */
+
+typedef struct {
```
keithc-ca, line 1722: Please give the `struct` a tag (`RSADP_BN`).

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1626,6 +1710,55 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSAEP
...
+/*
+ * Class:     jdk_crypto_jniprovider_NativeCrypto
+ * Method:    create_rsadp_BN
+ * Signature: ()J
+ */
+JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_create_1rsadp_1BN
+	(JNIEnv *env, jclass thisObj) {
+
+    RSADP_BN *rsadp_bn_ptr = (RSADP_BN *)malloc(sizeof(RSADP_BN));
+    rsadp_bn_ptr->ctx = (*OSSL_BN_CTX_new)();
```
keithc-ca, line 1738: This must check for `NULL`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -201,6 +256,17 @@ static void printErrors(void) {
...
+void OSSL_RSA_get0_key_compatibility(const RSA *r,
+                  const BIGNUM **n, const BIGNUM **e, const BIGNUM **d)
+ {
+    if (n != NULL)
+        *n = r->n;
+    if (e != NULL)
+        *e = r->e;
+    if (d != NULL)
+        *d = r->d;
+ }
```
keithc-ca, line 268: Opening and closing braces (`{}`) should be at the beginning of those lines.
Bodies of `if` statements should be enclosed in braces.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1626,6 +1710,55 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSAEP
...
+    return (jlong)rsadp_bn_ptr;
+}
+
+/*
+ * Class:     jdk_crypto_jniprovider_NativeCrypto
+ * Method:    free_rsadp_BN
+ * Signature: (J)V
+ */
+JNIEXPORT void JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_free_1rsadp_1BN
+	(JNIEnv *env, jclass thisObj, jlong rsadpBNptr) {
```
keithc-ca, line 1751: Opening brace on a line by itself.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1661,30 +1793,26 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSADP
...
+        if (verify == msg_len) {
+            /* mNative is size 'verify' */
+            /* verify by comparing BigNum of mNative and m_bigNum ^ e % n */
+            const BIGNUM *n = NULL;
+            const BIGNUM *e = NULL;
+            (*OSSL_RSA_get0_key)(rsaKey, (const BIGNUM **)&n, (const BIGNUM **)&e, NULL);
+            (*OSSL_BN_bin2bn)(kNative, kLen, rsadp_bn_ptr->k_bigNum);
+            (*OSSL_BN_bin2bn)(mNative, msg_len, rsadp_bn_ptr->m_bigNum);
+            if (NULL != rsadp_bn_ptr->k_bigNum && NULL != rsadp_bn_ptr->m_bigNum && NULL != e && NULL != n) {
+                if (!(*OSSL_BN_mod_exp)(rsadp_bn_ptr->rem, rsadp_bn_ptr->m_bigNum, e, n, rsadp_bn_ptr->ctx) || 0 != (*OSSL_BN_ucmp)(rsadp_bn_ptr->k_bigNum, rsadp_bn_ptr->rem)) {
```
keithc-ca, line 1805: Parentheses required, line excessively long:
```
    if (!(*OSSL_BN_mod_exp)(rsadp_bn_ptr->rem, rsadp_bn_ptr->m_bigNum, e, n, rsadp_bn_ptr->ctx)
        || (0 != (*OSSL_BN_ucmp)(rsadp_bn_ptr->k_bigNum, rsadp_bn_ptr->rem))
    ) {
```

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
```diff
@@ -59,6 +59,8 @@
 
     private static NativeCrypto nativeCrypto;
 
+    private static final ThreadLocal<RSADPBigNumRPointer> rsadpBnPtr = new ThreadLocal<RSADPBigNumRPointer>() {};
```
keithc-ca, line 62: Why an anonymous class here and not just this?
```
    private static final ThreadLocal<RSADPBigNumRPointer> rsadpBnPtr = new ThreadLocal<>();
```

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
```diff
@@ -138,17 +140,40 @@ public static int getByteLength(BigInteger b) {
...
         }
 
         return output;
     }
+
+    static class RSADPBigNumRPointer {
+        long rsadpBN = 0;
+
+        RSADPBigNumRPointer() {
+            this.rsadpBN = nativeCrypto.create_rsadp_BN();
```
keithc-ca, line 163: Using `finalize` is deprecated. Please use a `Cleaner`, like in `NativeDigest.java`.

keithc-ca (changes requested):

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
samasri: I removed this. The `synchronized` keyword was added here to signify that this function is not thread safe. So if multiple threads are calling the same object instance's `finalize` simultaneously, this would not be safe since the native function inside is freeing up a memory pointer.

However, after discussing this with Peter privately, he pointed out that the JVM would not invoke the `finalize` method for the same instance more than once. Hence, it is not necessary to include the `synchronized` keyword here.

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
samasri: That's right, I removed it.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
samasri: You mean like [this](https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/b39d79d05e1ad38d6c032589d17c9ad992a0585f/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c#L263-L271), right?

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
samasri: From the mentioned comment above, my understanding is that this opacity principle is added in OpenSSL 1.1.x. The structure I add here is for the JDK's compatibility with OpenSSL 1.0.x.

More specifically, the `rsa_st` defined here is needed to compile `OSSL_RSA_get0_key_compatibility`. The last is only used when the JDK cannot find the function `RSA_get0_key` ([corresponding source code](https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/8c626311828f8601f0be3165ee136d4393b0b82c/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c#L427-L429)). That function is found only in OpenSSL 1.1.x and not in OpenSSL 1.0.x. Hence, the only time we use `OSSL_RSA_get0_key_compatibility` is in OpenSSL 1.0.x.

So the opacity principle of OpenSSL 1.1.x is respected, but this definition here is needed to compile a function for backward compatibility.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
samasri: Just wanted to make sure I understand what you meant by tag before resolving this; [this](https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/f50abc2/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c#L1726) is what you meant right?

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
pshipton: We should also consider wrapping the RSADPBigNumRPointer in a SoftReference, so they don't stick around forever on all the threads. I see jdk.internal.ref.CleanerImpl.SoftCleanableRef() can be used for this.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
keithc-ca: Better, but there's still an extra space before the braces on lines 261 and 268.

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
samasri: @pshipton I added this soft reference in my most recent commit. Based on your suggestion, I also created [this test](https://gist.github.ibm.com/samasri/833b75b77affaa474e27c683f30c6ed7) to make sure the code is working as expected.

The test works as follows:
1. Calls the `NativeRSACore.sign()` function
2. Calls `System.gc()` 32 times (which is supposed to trigger the GC to work, this should effectively delete the soft reference from memory)
3. Calls `NativeRSACore.sign()` again

I also added a print statement in the JDK at the time of the creation of the `RSADPBigNumRPointer` (specifically, I added it below [this line](https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/89799461d95cf16481cabed5fc1e6dba3f2689ed/closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java#L150) inside the if statement). When running the test, I can see that this statement is printed twice, which proves that it does get deleted by the GC in the right circumstances. Furthermore, I tried removing the `System.gc()` for loop from the test, and the print statement that I added in the JDK only appeared once. This is additional proof that the soft pointer is created twice because of the calls of `System.gc()`, which triggers the garbage collector.

pshipton: Thanks, lgtm. If you haven't already, it's worth adding a print in the cleaner to ensure it's working to clean up. You'll need the 32 System.gc()'s to trigger the cleaning.

pshipton: Actually, to avoid recompiling the code, you can use a method tracepoint instead of a print to ensure RSADPBigNumRPointerCleanerRunnable.run() is happening.

samasri: I added a print statement in [this function](https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/89799461d95cf16481cabed5fc1e6dba3f2689ed/closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java#L197-L199) and I can confirm that I can see the message when I have `System.gc()` in the test. And I don't see the statement when `System.gc` is not in the test.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1626,6 +1714,59 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSAEP
...
+ * Class:     jdk_crypto_jniprovider_NativeCrypto
+ * Method:    create_rsadp_BN
+ * Signature: ()J
+ */
+JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_create_1rsadp_1BN
+	(JNIEnv *env, jclass thisObj) {
+
+    RSADP_BN *rsadp_bn_ptr = (RSADP_BN *)malloc(sizeof(RSADP_BN));
+    if(rsadp_bn_ptr != NULL) {
+        rsadp_bn_ptr->ctx = (*OSSL_BN_CTX_new)();
```
pshipton, line 1743: I expect these allocations can fail so need to be checked as well.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -99,8 +141,15 @@ typedef int OSSL_RSA_set0_factors_t(RSA *, BIGNUM *, BIGNUM *);
 typedef void OSSL_RSA_free_t (RSA *);
 typedef int OSSL_RSA_public_decrypt_t(int, const unsigned char *, unsigned char *, RSA *, int);
 typedef int OSSL_RSA_private_encrypt_t (int, const unsigned char *, unsigned char *, RSA *, int);
-typedef BIGNUM* OSSL_BN_bin2bn_t (const unsigned char *, int, BIGNUM *);
-typedef void OSSL_BN_set_negative_t (BIGNUM *, int);
+typedef int OSSL_RSA_get0_key_t(const RSA *, const BIGNUM **, const BIGNUM **, const BIGNUM **);;
```
pshipton, line 144: There is an extra ; here

pshipton:
@keithc-ca are all your concerns addressed?

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
```diff
@@ -24,21 +24,26 @@
...
- * (c) Copyright IBM Corp. 2018, 2019 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2021 All Rights Reserved
  * ===========================================================================
  */
 
 package sun.security.rsa;
 
 import java.math.BigInteger;
 import java.util.*;
+import java.lang.OutOfMemoryError;
```
keithc-ca, line 35: You don't need to import types from the `java.lang` package.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -99,8 +141,15 @@ typedef int OSSL_RSA_set0_factors_t(RSA *, BIGNUM *, BIGNUM *);
...
+typedef BIGNUM* OSSL_BN_bin2bn_t(const unsigned char *, int, BIGNUM *);
+typedef int OSSL_BN_mod_exp_t(BIGNUM *, BIGNUM *, const BIGNUM *, const BIGNUM *, BN_CTX *);
+typedef int OSSL_BN_ucmp_t(BIGNUM *, BIGNUM *);
+typedef int OSSL_BN_zero_t(BIGNUM *a);
+typedef void OSSL_BN_set_negative_t(BIGNUM *, int);
+typedef BN_CTX *OSSL_BN_CTX_new_t();
```
keithc-ca, line 150: Please consistently bind the `*` to the right (to reflect the meaning); write
```
typedef BIGNUM *OSSL_BN_bin2bn_t(const unsigned char *, int, BIGNUM *);
```
instead of
```
typedef BIGNUM* OSSL_BN_bin2bn_t(const unsigned char *, int, BIGNUM *);
```

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -347,10 +417,21 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
...
+    OSSL_BN_clear = (OSSL_BN_zero_t *)find_crypto_symbol(crypto_library, "BN_clear");
+    OSSL_BN_CTX_new = (OSSL_BN_CTX_new_t *)find_crypto_symbol(crypto_library, "BN_CTX_new");
+    OSSL_BN_CTX_free = (OSSL_BN_CTX_free_t *)find_crypto_symbol(crypto_library, "BN_CTX_free");
+    OSSL_BN_new = (OSSL_BN_new_t *)find_crypto_symbol(crypto_library, "BN_new");
     OSSL_BN_set_negative = (OSSL_BN_set_negative_t *)find_crypto_symbol(crypto_library, "BN_set_negative");
     OSSL_BN_free = (OSSL_BN_free_t *)find_crypto_symbol(crypto_library, "BN_free");
 
+    if (OSSL_RSA_get0_key == NULL) {
+        OSSL_RSA_get0_key = (OSSL_RSA_get0_key_t *) &OSSL_RSA_get0_key_compatibility;
+    }
```
keithc-ca, line 433: This should not apply for OpenSSL 1.1.*, right? If so, that condition should be included (i.e. `0 != ossl_ver`).

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1626,6 +1714,59 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSAEP
...
+JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_create_1rsadp_1BN
+	(JNIEnv *env, jclass thisObj) {
+
+    RSADP_BN *rsadp_bn_ptr = (RSADP_BN *)malloc(sizeof(RSADP_BN));
+    if (rsadp_bn_ptr != NULL) {
+        rsadp_bn_ptr->ctx = (*OSSL_BN_CTX_new)();
+        rsadp_bn_ptr->k_bigNum = (*OSSL_BN_new)();
+        rsadp_bn_ptr->m_bigNum = (*OSSL_BN_new)();
+        rsadp_bn_ptr->rem = (*OSSL_BN_new)();
+        return (jlong)rsadp_bn_ptr;
```
keithc-ca: Casts between `jlong` and pointers really ought to go through `intptr_t` to avoid changing signed-ness and size in one step.
* `jlong` to `T *` : `(T *)(intptr_t)longValue`
* `T *` to `jlong` : `(jlong)(intptr_t)somePointer`

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1626,6 +1714,59 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSAEP
...
+ * Method:    RSADP
+ * Signature: ([BI[BIJ)I
+ */
+
+typedef struct RSADP_BN {
+    BN_CTX *ctx;
+    BIGNUM *k_bigNum;
+    BIGNUM *m_bigNum;
+    BIGNUM *rem;
+} RSADP_BN;
```
keithc-ca, line 1731: The comment seems to be for a function rather than a type declaration; please fix.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1635,13 +1776,12 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSAEP
  * Signature: ([BI[BIJ)I
```
keithc-ca, line 1784: The signature has changed; please update the comment.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1661,30 +1801,28 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSADP
...
-                            break;
-                        }
-                    }
-                } else {
+        if (verify == msg_len) {
+            /* mNative is size 'verify' */
+            /* verify by comparing BigNum of mNative and m_bigNum ^ e % n */
+            const BIGNUM *n = NULL;
+            const BIGNUM *e = NULL;
+            (*OSSL_RSA_get0_key)(rsaKey, (const BIGNUM **)&n, (const BIGNUM **)&e, NULL);
```
keithc-ca, line 1817: The casts for the second and third arguments seem redundant.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -47,6 +48,47 @@
 # include <pthread.h>
 #endif /* defined(WINDOWS) */
 
+/*
+ * This definition is needed to compile the "OSSL_RSA_get0_key_compatibility" function
+ * Since the first parameter of that function is of type RSA, which is
+ * a typedef of rsa_st, the function would not compile without defining
+ * rsa_st here */
+struct rsa_st {
```
keithc-ca, line 56: I still object to this declaration. We should not assume anything about the definition of what is intended to be an opaque type.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
keithc-ca: If any allocation fails, the resources acquired via successful calls must be released.

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
```diff
@@ -138,17 +145,57 @@ public static int getByteLength(BigInteger b) {
...
             output = new byte[outputLen];
             int verifyInt = verify ? outputLen : -1;
-            outputLen = nativeCrypto.RSADP(msg, msg.length, output, verifyInt, nativePtr);
+            SoftReference<RSADPBigNumRPointer> rsadpBnSoftPtr = rsadpBnPtr.get();
+            RSADPBigNumRPointer rsadpBn = rsadpBnSoftPtr == null ? null : rsadpBnSoftPtr.get();
+            if (rsadpBn == null) {
+                try {
+                    rsadpBn = new RSADPBigNumRPointer();
+                }
+                catch(OutOfMemoryError e) {
```
keithc-ca, line 154: This should be on one line with a space before the opening parenthesis:
```
    } catch (OutOfMemoryError e) {
```

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
```diff
@@ -138,17 +145,57 @@ public static int getByteLength(BigInteger b) {
...
 
         return output;
     }
+
+    static class RSADPBigNumRPointer {
+
+        long rsadpBN = 0;
+        private static final Cleaner rsadpPtrCleaner = CleanerFactory.cleaner();
+
+        RSADPBigNumRPointer() throws OutOfMemoryError {
```
keithc-ca, line 176: Where are the strong references to instances of this class? I see none, in which case they may be collected prematurely leaving a dangling reference to the native memory that was release by the cleaner thread.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
keithc-ca: Further, this appears to be a copy of https://github.com/openssl/openssl/blob/OpenSSL_1_1_1k/crypto/rsa/rsa_local.h#L28-L66, but I see no mention of the openssl licenses in this file.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
keithc-ca: Correction: `OSSL_RSA_get0_key_compatibility` should only be used if `0 == ossl_ver`.

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
samasri: @keithc-ca Yes, that is correct. When we reach the `reachabilityFence`, we already have all the needed data and the native `RSADPBigNumRPointer` can be safely reclaimed.

Sure, I changed the code to call `create_rsadp_BN()` before constructing the `RSADPBigNumRPointer`.

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
```diff
@@ -138,17 +145,57 @@ public static int getByteLength(BigInteger b) {
...
+                }
+                catch(OutOfMemoryError e) {
+                    return null;
+                }
+                rsadpBnPtr.set(new SoftReference<>(rsadpBn));
+            }
+            outputLen = nativeCrypto.RSADP(msg, msg.length, output, verifyInt, nativePtr, rsadpBn.getBN());
+            Reference.reachabilityFence(rsadpBn);
         } finally {
             key.returnNativePtr(nativePtr);
```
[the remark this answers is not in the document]
samasri: That is correct.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
samasri: @keithc-ca The issue is, I don't think there's a way around it. I've been trying to look at other functions in OpenSSL 1.0.* that would return the same fields as `RSA_get0_key` in OpenSSL 1.1.*, but I couldn't find any. And this makes sense because `rsa_st` was not opaque and those fields could be accessed directly without the need for functions.

In [OpenSSL's wiki](https://wiki.openssl.org/index.php/OpenSSL_1.1.0_Changes), it is mentioned that one of the API changes in 1.1.0 as compared to previous versions is that the structs are now opaque. The backward compatibility issue was also expected and the the suggested solution for that was to make those compatibility functions. Now if we compile the JDK with OpenSSL 1.0.*, we won't need to add this structure definition since it is available in an included header file ([rsa.h](https://github.com/openssl/openssl/blob/OpenSSL_1_0_0-stable/crypto/rsa/rsa.h#L132-L165)). However, in OpenSSL 1.1.*, the structure is not defined in any included header file and is not expected to be used. Hence, the only way that I could think of to support OpenSSL 1.0.* even when compiling the JDK with OpenSSL 1.1.* is to add the definition itself there.

In terms of the `lock` field, you're right. I took that from OpenSSL 1.1.1 by mistake. I removed it in my latest commit

AdamBrousseau:
Jenkins line endings check

AdamBrousseau:
Jenkins copyright check

AdamBrousseau:
jenkins compile osx jdk11

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
samasri: After discussing this more with Keith and Alon on slack, I solve this issue as follows:
I remove the `rsa_st` definition from our native source code and fall back on the pure java implementation when necessary.

In pseudocode:

```java
if (openssl version >= 1.1.*) {
  keep using our new code to verify
} else {
  if (key has a size that would cause an error in the old native implementation) {
    use pure java implementation
  } else {
    use OSSL_RSA_public_decrypt to verify
  }
}
```

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
```diff
@@ -138,17 +144,58 @@ public static int getByteLength(BigInteger b) {
...
         if (outputLen == -1) {
-            return null;
-        } else if (outputLen == -2) {
             throw new BadPaddingException("RSA private key operation failed");
         }
 
         return output;
     }
+
+    static class RSADPBigNumRPointer {
```
keithc-ca, line 174: This class should be `private` and `final`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -389,7 +411,14 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
         (NULL == OSSL_RSA_free) ||
         (NULL == OSSL_RSA_public_decrypt) ||
         (NULL == OSSL_RSA_private_encrypt) ||
+        ((NULL == OSSL_RSA_get0_key) && ossl_ver == 1) || /* RSA_get0_key only exists in OpenSSL >= 1.1.1 */
```
keithc-ca, line 414: Missing parentheses: `(ossl_ver == 1)`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -347,7 +362,14 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
     OSSL_RSA_free = (OSSL_RSA_free_t *)find_crypto_symbol(crypto_library, "RSA_free");
     OSSL_RSA_public_decrypt = (OSSL_RSA_public_decrypt_t *)find_crypto_symbol(crypto_library, "RSA_public_decrypt");
     OSSL_RSA_private_encrypt = (OSSL_RSA_private_encrypt_t *)find_crypto_symbol(crypto_library, "RSA_private_decrypt");
+    OSSL_RSA_get0_key = (OSSL_RSA_get0_key_t *)find_crypto_symbol(crypto_library, "RSA_get0_key");
```
keithc-ca, line 365: We should not attempt this lookup if `(ossl_ver == 0)`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1626,17 +1655,90 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSAEP
...
+    BIGNUM *rem;
+} RSADP_BN;
+
+/*
+ * Class:     jdk_crypto_jniprovider_NativeCrypto
+ * Method:    create_rsadp_BN
+ * Signature: ()J
+ */
+JNIEXPORT jlong JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_create_1rsadp_1BN
+	(JNIEnv *env, jclass thisObj) {
```
keithc-ca, line 1672: Formatting:
```
JNIEXPORT jlong JNICALL
Java_jdk_crypto_jniprovider_NativeCrypto_create_1rsadp_1BN(JNIEnv *env, jclass thisObj)
{
```

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1626,17 +1655,90 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSAEP
...
+
+    RSADP_BN *rsadp_bn_ptr = (RSADP_BN *)malloc(sizeof(RSADP_BN));
+    if (rsadp_bn_ptr == NULL) {
+        return (jlong)(intptr_t) -1;
```
keithc-ca, line 1676: There's no need to cast through `intptr_t` here; `(jlong)-1` is sufficient.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1626,17 +1655,90 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSAEP
...
+	(JNIEnv *env, jclass thisObj, jlong rsadpBNptr)
+{
+
+    RSADP_BN *rsadp_bn_ptr = (RSADP_BN *)rsadpBNptr;
+    (*OSSL_BN_CTX_free)(rsadp_bn_ptr->ctx);
+    (*OSSL_BN_free)(rsadp_bn_ptr->k_bigNum);
+    (*OSSL_BN_free)(rsadp_bn_ptr->m_bigNum);
+    (*OSSL_BN_free)(rsadp_bn_ptr->rem);
+    free(rsadp_bn_ptr);
+    return;
```
keithc-ca, line 1727: Please remove this redundant `return` statement.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1661,30 +1763,55 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSADP
...
-                msg_len2 = (*OSSL_RSA_public_decrypt)(verify, mNative, k2, rsaKey, RSA_NO_PADDING);
-                if (-1 != msg_len2) {
-
-                    int i;
-                    for (i = 0; i < verify; i++) {
-                        if (kNative[i] != k2[i]) {
-                            msg_len = -2;
-                            break;
+        if (verify == msg_len) {
+            if(-1 != rsadpBNptr) {
```
keithc-ca, line 1767: Space after `if`.
Why did you choose `-1` instead of `0` as the special value?

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1661,30 +1763,55 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSADP
...
+                rsadp_bn_ptr = (RSADP_BN *)rsadpBNptr;
+                /* We are using OpenSSL 1.1.*
+                 * Verify by comparing BigNum of mNative and m_bigNum ^ e % n */
+                const BIGNUM *n = NULL;
+                const BIGNUM *e = NULL;
+                (*OSSL_RSA_get0_key)(rsaKey, &n, &e, NULL);
+                (*OSSL_BN_bin2bn)(kNative, kLen, rsadp_bn_ptr->k_bigNum);
+                (*OSSL_BN_bin2bn)(mNative, msg_len, rsadp_bn_ptr->m_bigNum);
+                if (NULL != rsadp_bn_ptr->k_bigNum && NULL != rsadp_bn_ptr->m_bigNum && NULL != e && NULL != n) {
```
keithc-ca, line 1776: Four sets of missing parentheses.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1661,30 +1763,55 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSADP
...
+                        msg_len = -1;
+                    }
+                } else {
+                    msg_len = -1;
+                }
+                (*OSSL_BN_clear)(rsadp_bn_ptr->k_bigNum);
+                (*OSSL_BN_clear)(rsadp_bn_ptr->m_bigNum);
+                (*OSSL_BN_clear)(rsadp_bn_ptr->rem);
+            }
+            else {
```
keithc-ca, line 1789: `} else {`

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1661,30 +1763,55 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSADP
...
+                */
+                k2 = malloc(kLen * (sizeof(unsigned char)));
+                if (NULL != k2) {
+                    /* mNative is size 'verify' */
+                    msg_len2 = (*OSSL_RSA_public_decrypt)(verify, mNative, k2, rsaKey, RSA_NO_PADDING);
+                    if (-1 != msg_len2) {
+                        int i;
+                        for (i = 0; i < verify; i++) {
+                            if (kNative[i] != k2[i]) {
+                                msg_len = -2;
```
keithc-ca, line 1801: The java code no longer handles the return value `-2`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1661,30 +1763,55 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSADP
...
+                (*OSSL_BN_clear)(rsadp_bn_ptr->rem);
+            }
+            else {
+                /* We are using OpenSSL 1.0.*
+                 * Verify by comparing RSA_public_decrypt with original msg
+                */
+                k2 = malloc(kLen * (sizeof(unsigned char)));
+                if (NULL != k2) {
+                    /* mNative is size 'verify' */
+                    msg_len2 = (*OSSL_RSA_public_decrypt)(verify, mNative, k2, rsaKey, RSA_NO_PADDING);
```
keithc-ca, line 1796: Why doesn't this approach also work for OpenSSL 1.1.*?

keithc-ca:
Please also update the copyright dates.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
samasri: You meant the ones in the if statement right? So it should be like this: `if ((NULL != rsadp_bn_ptr->k_bigNum) && (NULL != rsadp_bn_ptr->m_bigNum) && (NULL != e) && (NULL != n)) {`

src/java.base/share/classes/sun/security/rsa/RSACore.java
```diff
@@ -145,7 +152,7 @@ public static int getByteLength(RSAKey key) {
     public static byte[] rsa(byte[] msg, RSAPrivateKey key, boolean verify)
             throws BadPaddingException {
         if (key instanceof RSAPrivateCrtKey) {
-            if (useNativeRsa && key instanceof sun.security.rsa.RSAPrivateCrtKeyImpl) {
+            if (useNativeRsa && key instanceof sun.security.rsa.RSAPrivateCrtKeyImpl && (NativeCrypto.getVersion() == 0 && !unsupportedKeySizes.contains(key.getModulus().bitLength()))) {
```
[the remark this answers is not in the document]
keithc-ca: That comment just points back here, and I couldn't find the answer here either. Did I just miss the explanation?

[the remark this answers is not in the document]
samasri: I should've pointed to the issue instead of the PR. I updated the comment to point to the issue and added a better summary of the issue and how its solved in [a comment[(https://github.com/eclipse-openj9/openj9/issues/11500#issuecomment-833004636)

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
```diff
@@ -138,17 +144,58 @@ public static int getByteLength(BigInteger b) {
             outputLen = getByteLength(n);
             output = new byte[outputLen];
             int verifyInt = verify ? outputLen : -1;
-            outputLen = nativeCrypto.RSADP(msg, msg.length, output, verifyInt, nativePtr);
+            if(nativeCrypto.getVersion() == 0) {
+                outputLen = nativeCrypto.RSADP(msg, msg.length, output, verifyInt, nativePtr, -1);
+            }
+            else {
```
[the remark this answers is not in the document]
keithc-ca: You didn't fix this. Please leave me to mark conversations as resolved so I can keep track of the outstanding points.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
keithc-ca: Zero is slightly easier to deal with and it's the natural choice for a pointer.

closed/src/java.base/share/classes/sun/security/rsa/NativeRSACore.java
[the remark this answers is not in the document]
samasri: Okay, I fixed it in the latest commit.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
keithc-ca: Yes.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
keithc-ca: I still don't understand why 1025 is special. Please explain, and make sure the answer is in the code somewhere for easy future reference.

src/java.base/share/classes/sun/security/rsa/RSACore.java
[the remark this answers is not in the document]
keithc-ca: The new comment is better, but what is still not clear is why 1025 is problematic. Is it because it's not a power of 2 (so 1023, 2047 and 2049 would also be padded)? If so, the code should be corrected to properly identify when padding will occur; or better, be independent of openssl implementation details.

src/java.base/share/classes/sun/security/rsa/RSACore.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2003, 2020, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2003, 2021, Oracle and/or its affiliates. All rights reserved.
```
keithc-ca, line 2: This copyright date should not change; you need to change the one below on line 27.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
samasri: Okay, changed it to zero.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
samasri: Okay, then this is fixed now.

samasri: Okay, I changed that.

src/java.base/share/classes/sun/security/rsa/RSACore.java
```diff
@@ -137,6 +137,24 @@ public static int getByteLength(RSAKey key) {
...
+     * it always pads the result
+     * The key sizes that will cause padding issues are the ones of size: 8n+1
+     * (where n is any whole number)
+     * Hence, this function returns false if the key size is can be expressed
+     * in the form of 8n+1
+     * 
+     * @param keySize The size of the key that needs to be checked
+     * @return        False if the key can be expressed as 8n+1, True otherwise
+     * See https://github.com/eclipse-openj9/openj9/issues/11500#issuecomment-833004636 for more info
+    */
```
keithc-ca, line 153: Please finish each sentence with a period. The final line of the comment should be indented one more space.

src/java.base/share/classes/sun/security/rsa/RSACore.java
```diff
@@ -137,6 +137,24 @@ public static int getByteLength(RSAKey key) {
...
+    private static boolean checkKeySizeSupport(int keySize) {
+        return ((keySize - 1) % 8) != 0;
```
keithc-ca, line 155: How about `(keySize % 8) != 1`?

src/java.base/share/classes/sun/security/rsa/RSACore.java
```diff
@@ -145,7 +163,9 @@ public static int getByteLength(RSAKey key) {
     public static byte[] rsa(byte[] msg, RSAPrivateKey key, boolean verify)
             throws BadPaddingException {
         if (key instanceof RSAPrivateCrtKey) {
-            if (useNativeRsa && key instanceof sun.security.rsa.RSAPrivateCrtKeyImpl) {
+            if (useNativeRsa
+                && key instanceof sun.security.rsa.RSAPrivateCrtKeyImpl
+                && ((NativeCrypto.getVersion() != 0) || (checkKeySizeSupport(key.getModulus().bitLength())))) {
```
keithc-ca, line 168: Redundant parentheses in `(checkKeySizeSupport(key.getModulus().bitLength()))`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
samasri: The test is only failing when the key size is `8n + 1` (where n is a whole number). I tried running the test (when triggering OpenSSL's `RSA_public_decrypt`) on key sizes: 1023, 1024, 1025, and 1026; the test is only failing on 1025.

Hence, I updated the code to check for that pattern (`keySize % 8 == 1`).

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -389,7 +413,14 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_loadCrypto
         (NULL == OSSL_RSA_free) ||
         (NULL == OSSL_RSA_public_decrypt) ||
         (NULL == OSSL_RSA_private_encrypt) ||
+        ((NULL == OSSL_RSA_get0_key) && (ossl_ver == 1)) || /* RSA_get0_key only exists in OpenSSL >= 1.1.1 */
```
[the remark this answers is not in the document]
samasri: Okay, I flipped those conditions.

src/java.base/share/classes/sun/security/rsa/RSACore.java
samasri: Done.

src/java.base/share/classes/sun/security/rsa/RSACore.java
samasri: Done.

src/java.base/share/classes/sun/security/rsa/RSACore.java
samasri: Okay, I removed them.

src/java.base/share/classes/sun/security/rsa/RSACore.java
samasri: Fixed that.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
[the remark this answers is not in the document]
keithc-ca: Constants on the left of `==` or `!=`:
```
        ((1 == ossl_ver) && (NULL == OSSL_RSA_get0_key)) || /* RSA_get0_key only exists in OpenSSL >= 1.1.1 */
```

src/java.base/share/classes/sun/security/rsa/RSACore.java
keithc-ca: The comment doesn't reflect the actual use pattern. If the method returns `false`, OpenSSL 1.0.x is still used, but a different verification approach is used.

samasri:
There are two ways to fix the issue corresponding to this PR:
* **Approach 1:** Replace the `RSA_public_decrypt` with other OpenSSL functions. Those functions are not part of its public API, but are exported in the built _so_ file so they can be imported and used relatively easily.
  * Pro: it is a single implementation that works for all cases.
  * Con: we need to add [a compatibility layer](https://wiki.openssl.org/index.php/OpenSSL_1.1.0_Changes#Compatibility_Layer) since one of the functions is not found in previous versions of OpenSSL (versions 1.0.x). For that compatibility layer to compile, we need to copy one structure definition from the OpenSSL source code.
* **Approach 2:** Keep using the OpenSSL function (`RSA_public_decrypt`) that is currently used and alter its output when necessary.
  * Pro: This function is part of the public OpenSSL API.
  * Con: We're basically getting around a weird behaviour in OpenSSL by altering the output.

After discussing this issue more with Alon and Keith, we decided to go with the second approach. Hence, my I forced pushed onto the branch to replace the previous implementation (which was the first approach) with new implementation (which is the second approach).

keithc-ca (changes requested):
I think this should be rebased on the head of the openj9 branch and squashed to a single commit. The message in that commit should be more descriptive than either
* Handling exceptional cases
* Updated copyright and wording

Likewise, the description of this PR should be updated.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1670,10 +1670,25 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSADP
...
+                     */
+                    if (verify == kLen + 1) {
+                        for (i = 0; i < kLen; i++) {
+                            if (kNative[i] != k2[i + 1]) {
+                                msg_len = -2;
+                                break;
+                            }
+                        }
+                    }
+                    else { // if verify == kLen
```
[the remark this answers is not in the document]
samasri: Done.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1661,7 +1661,7 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSADP
     msg_len = (*OSSL_RSA_private_encrypt)(kLen, kNative, mNative, rsaKey, RSA_NO_PADDING);
 
     if ((-1 != verify) && (-1 != msg_len)) {
-        if (verify == kLen) {
+        if (verify == kLen || verify == kLen + 1) {
```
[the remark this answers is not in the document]
samasri: Done.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1670,10 +1670,25 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSADP
...
-                    for (i = 0; i < verify; i++) {
-                        if (kNative[i] != k2[i]) {
-                            msg_len = -2;
-                            break;
+                    /*
+                     * For certain key sizes, the decrypted message retrieved from the RSA_public_decrypt
+                     * includes a 1 byte padding at the beginning of the message. In these cases, this
+                     * padding must be ignored when comparing to the original message.
+                     */
+                    if (verify == kLen + 1) {
```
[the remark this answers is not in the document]
samasri: Done.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1670,10 +1670,25 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSADP
...
+                        for (i = 0; i < kLen; i++) {
```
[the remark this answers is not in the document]
samasri: True, fixed it.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1670,10 +1670,28 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSADP
...
                     int i;
-                    for (i = 0; i < verify; i++) {
-                        if (kNative[i] != k2[i]) {
+                    /*
+                     * For certain key sizes, the decrypted message retrieved from the RSA_public_decrypt
+                     * includes a 1 byte padding at the beginning of the message. In these cases, this
+                     * padding must be ignored when comparing to the original message.
+                     */
+                    if (verify == (kLen + 1)) {
+                        if (k2[0] != 0) {
```
keithc-ca, line 1679: Constant on the left, please: `0 != k2[0]`.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1670,10 +1670,28 @@ JNIEXPORT jint JNICALL Java_jdk_crypto_jniprovider_NativeCrypto_RSADP
                 if (-1 != msg_len2) {
 
                     int i;
-                    for (i = 0; i < verify; i++) {
-                        if (kNative[i] != k2[i]) {
+                    /*
+                     * For certain key sizes, the decrypted message retrieved from the RSA_public_decrypt
+                     * includes a 1 byte padding at the beginning of the message. In these cases, this
+                     * padding must be ignored when comparing to the original message.
```
keithc-ca, line 1676: Please correct this comment. We're not ignoring that padding; instead, we expect it to be a zero byte.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
samasri: Done.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
samasri: Right, fixed it.

keithc-ca:
I think I'm happy with the code now.

I'd like to see this squashed, and I still think the commit message can be improved.

The first (summary) line should start with a verb in the present tense ('Handle exceptional ...'). The title of this pull request should follow the same pattern. There's no limit to the size of the remainder of the commit message, so there's need to omit important details like the motivation for the change. There's some good advice at https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#commit-guidelines along with links to several other sites on that topic.

You'll also note that I updated the description so this does not close the openj9 issue: that should not occur until this has been propagated to all affected repositories.

samasri:
Sure, done. Thanks for all the revisions!

keithc-ca (approved):

keithc-ca:
Jenkins test openjdk.extended plinux jdk11

pshipton:
We should use grinders for the 2 tests listed in https://github.com/eclipse-openj9/openj9/issues/11500, they aren't part of sanity.openjdk. I can do that once the JVM build is ready.

jenkins compile win,zlinux jdk11

keithc-ca:
Did I misread the aqa-tests [playlist](https://github.com/adoptium/aqa-tests/blob/master/openjdk/playlist.xml#L677)? It lists `jdk_security3` which I understood to include the relevant tests (in openjdk.extended).

pshipton:
You can see a run of extended.openjdk I did yesterday, and SignatureTestPSS, SignatureTest2 are not listed.https://openj9-jenkins.osuosl.org/job/Test_openjdk11_j9_extended.openjdk_ppc64_aix_Personal/1/testReport/

pshipton:
They may be excluded/disabled. Also since the test is called extended.openjdk, not sure if openjdk.extended is going to run anything.

Plus we should run sanity.openjdk to ensure this change doesn't regress anything there.

pshipton:
sanity.openjdk - https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/48/
SignatureTestPSS - https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/49/
SignatureTest2 - https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/50/

keithc-ca:
Jenkins test sanity,sanity.openjdk osx jdk11

pshipton:
The Mac failure is infra, half the tests passed, and the other half failed to run due to the following, which affected the nightly builds as well.
```
Also:   hudson.model.Computer$TerminationRequest: Termination requested at Mon May 31 20:07:38 EDT 2021 by Thread[Running CpsFlowExecution[Owner[Sanitize-Nodes/13:Sanitize-Nodes #13]],5,main] [id=18861]
```

pshipton (approved):

PR #405, opened 2021-05-07
Allow more flexible compiler invocations on z/OS

dnakamura:
Allow compiler options to follow input file names. This is needed for
CMake builds.

Signed-off-by: Devin Nakamura <devinn@ca.ibm.com>

PR #406, opened 2021-05-07
Always define EXPORT_COMPILER_ENV_VARS

keithc-ca:
#405 missed renaming `EXPORT_MSVS_ENV_VARS` in the default case.


keithc-ca:
FYI @dnakamura

pshipton (approved):

PR #408, opened 2021-05-20
In test, increase deflate buffer size for zlinux Ubuntu 20

pshipton:
test/jdk/java/util/zip/DeInflate.java

Issue https://github.com/eclipse-openj9/openj9/issues/12740

Tested via https://ci.eclipse.org/openj9/view/Test/job/Grinder/1836/ - passed

Note using a smaller size of + 16K isn't big enough.
https://ci.eclipse.org/openj9/view/Test/job/Grinder/1835/
```
12:15:58  m=540672, n=512675, len=524288, eq=false
12:15:58  STDERR:
12:15:58  java.lang.RuntimeException: De/inflater failed:java.util.zip.Deflater@a9ab8748
12:15:58  	at DeInflate.check(DeInflate.java:146)
12:15:58  	at DeInflate.main(DeInflate.java:295)
```

### DIFF
--- a/test/jdk/java/util/zip/DeInflate.java
+++ b/test/jdk/java/util/zip/DeInflate.java
@@ -20,6 +20,11 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2021, 2021 All Rights Reserved
+ * ===========================================================================
+ */
 
 /**
  * @test
@@ -268,7 +273,7 @@ public class DeInflate {
 
         byte[] dataIn = new byte[1024 * 512];
         rnd.nextBytes(dataIn);
-        byte[] dataOut1 = new byte[dataIn.length + 1024];
+        byte[] dataOut1 = new byte[dataIn.length + (32 * 1024)]; // See https://github.com/eclipse-openj9/openj9/issues/12740
         byte[] dataOut2 = new byte[dataIn.length];
 
         Deflater defNotWrap = new Deflater(Deflater.DEFAULT_COMPRESSION, false);


pshipton:
Added a comment to the code after testing the change.

test/jdk/java/util/zip/DeInflate.java
```diff
@@ -268,7 +273,7 @@ public static void main(String[] args) throws Throwable {
 
         byte[] dataIn = new byte[1024 * 512];
         rnd.nextBytes(dataIn);
-        byte[] dataOut1 = new byte[dataIn.length + 1024];
+        byte[] dataOut1 = new byte[dataIn.length + (32 * 1024)]; // See https://github.com/eclipse-openj9/openj9/issues/12740
```
keithc-ca, line 276: 12.5% extra seems reasonable, but do we know what the maximum increase in size might be for the deflate algorithm?

keithc-ca:
Please update your commit with the 'signed-off-by' line.

test/jdk/java/util/zip/DeInflate.java
pshipton: Note deflate is supposed to make it smaller.

pshipton:
jenkins signed-off-by check

test/jdk/java/util/zip/DeInflate.java
keithc-ca: No compression algorithm can guarantee to make it smaller. The test appears to operate on random input so we should not be surprised that sometimes the 'compressed' data is larger, the question is how much larger?

pshipton:
jenkins compile osx jdk11

AdamBrousseau:
Jenkins copyright check

test/jdk/java/util/zip/DeInflate.java
pshipton: I tried various values, and the failing test showed how much data was re-inflated. With 16K it re-inflated 512675 (out of 524288), and with 8K extra it re-inflated 50494. The test runs twice with different data, and in both runs the numbers were similar though not exactly the same. Around + 28K should be good enough, but I allow for 32K.

pshipton: Lost a digit there, 8K was 504941 not 50494.

keithc-ca:
I assume jdk16 and jdknext should have similar fixes?
Should we arrange to propose this change upstream?

pshipton:
I'll create changes for jdk16 and next, and check java 8 as well.

> Should we arrange to propose this change upstream?

Not sure yet. It may be specific to zlinux, we'd need to try other platforms of Ubuntu 20. It seems more of a libz.so problem, previously it only needed 1K extra and now it needs ~28K.

keithc-ca (approved):

pshipton:
The work around seems unneeded on jdk8, the test doesn't fail there.
https://openj9-jenkins.osuosl.org/view/Test/job/Grinder/3/

PR #410, opened 2021-06-01
Ensure healthcenter.properties file is properly encoded on z/OS

keithc-ca:
Otherwise, `sed` will not properly interpret its contents.


pshipton:
I'd prefer for the end result to be the healthcenter.properties file in EBCDIC, rather than tagged ASCII. This matches the other properties files, and with IBM Java 8.

keithc-ca:
They're ASCII, but not tagged, just the way the are in the source archive.
If we're going to convert to EBCDIC and tag them accordingly, I think there are other properties files that should get the same treatment. I'll investigate what's on that list.

pshipton:
The other properties files I found are untagged EBCDIC, I don't think EBCDIC needs to be tagged.

pshipton:
jenkins compile zlinux jdk11

pshipton (approved):

PR #440, opened 2021-08-25
Add openj9.traceformat to the list of platform modules

keithc-ca:
It is required by ibm.healthcenter: a platform module cannot require non-platform modules.


pshipton:
jenkins compile zlinux jdk11

PR #450, opened 2021-10-27
Add support for building for AArch64 macOS

knn-k:
This commit adds the support for building for AArch64 macOS.

Signed-off-by: KONNO Kazuhiro <konno@jp.ibm.com>

### DIFF
--- a/closed/autoconf/custom-hook.m4
+++ b/closed/autoconf/custom-hook.m4
@@ -63,7 +63,7 @@ AC_DEFUN([OPENJ9_CONFIGURE_CMAKE],
     ],
     [
       case "$OPENJ9_PLATFORM_CODE" in
-        ap64|mz64|oa64|wa64|xa64|xl64|xr64|xz64)
+        ap64|mz64|oa64|or64|wa64|xa64|xl64|xr64|xz64)
           if test "x$COMPILE_TYPE" != xcross ; then
             with_cmake=cmake
           else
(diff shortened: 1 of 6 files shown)


0xdaryl:
@keithc-ca : please re-review

make/autoconf/build-aux/autoconf-config.guess
```diff
@@ -1270,6 +1270,7 @@ EOF
 		      UNAME_PROCESSOR="x86_64"
 		  fi
 		fi ;;
+	    arm) UNAME_PROCESSOR=aarch64 ;;
```
keithc-ca, line 1273: This was not needed for jdk17+: why is it needed here?

knn-k: I had to add it because the `configure` in JDK 11 detects the target processor as 32-bit ARM instead of AArch64.
I don't see the problem with JDK 17.

knn-k: I don't know what JDK 17's `configure` does for detecting the target processor on macOS.

knn-k: OK, I changed my commit to detect the target processor on macOS in the same way as JDK 17 does, adding some lines to make/autoconf/build-aux/config.guess instead of autoconf-config.guess.

keithc-ca: That change to `config.guess` appears to have originated from Azul, and so, I think we should also include the copyright addition from https://github.com/ibmruntimes/openj9-openjdk-jdk17/commit/e5ed92f29af6833890c1d511b5120e7e71bbf0c5#diff-93f22bc9c1a02cab4846b6c121a1ef46b9451060ec5fb84f7e872a2b8c23dce4.

knn-k: I added the Azul copyright line to config.guess.

keithc-ca:
Jenkins test sanity alinux,osx jdk11

keithc-ca:
Jenkins test sanity aarch64_linux jdk11

keithc-ca:
Jenkins test sanity osx jdk11

keithc-ca (approved):

PR #469, opened 2022-01-17
Set native stack size using -Xmso instead of -Xss

mikezhang1234567890:
OpenJ9 uses -Xmso to set native stack size, so launcher should also use this
value when creating the JVM thread.

Signed-off-by: Mike Zhang <mike.h.zhang@ibm.com>

Port of https://github.com/ibmruntimes/openj9-openjdk-jdk17/pull/57 to JDK11
Partially addresses https://github.com/eclipse-openj9/openj9/issues/12785

### DIFF
--- a/src/java.base/share/native/libjli/java.c
+++ b/src/java.base/share/native/libjli/java.c
@@ -24,6 +24,12 @@
  */
 
 /*
+ * ===========================================================================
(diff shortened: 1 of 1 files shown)


keithc-ca (approved):

keithc-ca:
jenkins test sanity osx jdk11

PR #474, opened 2022-01-27
Enable JIT on AArch64 macOS

knn-k:
This commit enables JIT on AArch64 macOS.

Signed-off-by: KONNO Kazuhiro <konno@jp.ibm.com>

PR #475, opened 2022-01-27
Remove a wrong copyright line from flags.m4

knn-k:
This commit removes a copyright line from flags.m4 that was added by
mistake in #474.

Signed-off-by: KONNO Kazuhiro <konno@jp.ibm.com>


keithc-ca (approved):

keithc-ca:
Thank you, @knn-k.

PR #484, opened 2022-03-07
Use bundled freetype on AIX

pshipton:
Backport https://github.com/ibmruntimes/openj9-openjdk-jdk17/commit/1e90c2e078

When libfreetype.so is bundled, users don't need to install it in order
to load libfontmanager.so


pshipton:
jenkins test sanity.openjdk aix jdk11

keithc-ca (approved):

PR #489, opened 2022-03-24
healthcenter: don't make jdk.internal.jvmstat a platform module

keithc-ca:
A build configured with `--with-healthcenter` requires that `ibm.healthcenter`, and the modules it requires, be platform modules.

- update the default security policy with permissions required when `jdk.attach` is a platform module
  - fixes a test failure in `CheckAccessClassInPackagePermissions`
- remove the dependency on `jdk.internal.jvmstat` so it need not be a platform module
- add stubs in `jdk.attach` for classes that used to depend on `jdk.internal.jvmstat`

### DIFF
--- a/closed/custom/common/Modules.gmk
+++ b/closed/custom/common/Modules.gmk
@@ -38,7 +38,7 @@ else
   PLATFORM_MODULES += \
 	ibm.healthcenter \
 	jdk.attach \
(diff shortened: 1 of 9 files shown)


pshipton:
jenkins compile amac jdk11

pshipton (approved):

PR #492, opened 2022-03-29
Stop ignoring warnings in openj9.sharedclasses

keithc-ca:
Depends on eclipse-openj9/openj9#14814.


pshipton:
jenkins compile amac jdk11

PR #494, opened 2022-03-30
Add CRIU security provider

ZainabF92:
Specialized security provider needed for CRIU

Signed-off-by: Zainab Fatmi <zainab@ibm.com>

PR #501, opened 2022-04-13
Apply JPP to JCL patch files within $(TOPDIR)/closed

JasonFengJ9:
Removed `$(TOPDIR)/closed/src` from `TOP_SRC_DIRS`;
Updated `jpp_generate_sources` for various configurations.

### DIFF
--- a/closed/GensrcJ9JCL.gmk
+++ b/closed/GensrcJ9JCL.gmk
@@ -29,12 +29,11 @@ J9TOOLS_DIR := $(SUPPORT_OUTPUTDIR)/j9tools
 JPP_JAR     := $(J9TOOLS_DIR)/jpp.jar
 
 RecursiveWildcard = $(foreach dir,$(wildcard $1/*),$(call RecursiveWildcard,$(dir),$2) $(filter $(subst *,%,$2),$(dir)))
-AllJclSource = $(call RecursiveWildcard,$(OPENJ9_TOPDIR)/jcl/src,*.java)
+JppSourceDirs := $(OPENJ9_TOPDIR)/jcl/src
+JppSourceDirs += $(TOPDIR)/closed/src
 
 ifeq (true,$(OPENJ9_ENABLE_DDR))
-  AllDdrSource = $(call RecursiveWildcard,$(OPENJ9_TOPDIR)/debugtools/DDR_VM/src,*.java)
-else
-  AllDdrSource =
+  JppSourceDirs += $(OPENJ9_TOPDIR)/debugtools/DDR_VM/src
 endif # OPENJ9_ENABLE_DDR
 
 JPP_TAGS := PLATFORM-$(OPENJ9_PLATFORM_CODE)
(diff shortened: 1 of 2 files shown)


closed/GensrcJ9JCL.gmk
```diff
@@ -79,6 +79,7 @@ $(J9JCL_SOURCES_DONEFILE) : $(AllJclSource) $(AllDdrSource) $(OPENJDK_STAGED_PRO
 		preprocessor
 	@$(ECHO) Generating J9JCL sources
 	$(call jpp_generate_sources,$(call FixPath,$(OPENJ9_TOPDIR)),jcl)
+	$(call jpp_generate_sources,$(call FixPath,$(TOPDIR)),closed)
```
keithc-ca, line 82: This is not enough by itself:

* There are missing dependencies on the source files within the `closed` folder - changing any of them will not trigger preprocessing again, but it should.
* `$(TOPDIR)/closed/src` should be removed from `TOP_SRC_DIRS` in `Modules.gmk`.

The order should be `overlay`, `closed`, `jcl` to retain the precedence of `jcl` over `closed` source.

I think we should use JPP options `-includeIfUnsure -noWarnIncludeIf` rather than modify all the files in `$(TOPDIR)/closed/src`.

JasonFengJ9: * Created `AllClosedSource` as prerequisite of `$(J9JCL_SOURCES_DONEFILE)`;
* Removed `$(TOPDIR)/closed/src` from `TOP_SRC_DIRS`;
* Reordered three `jpp_generate_sources` calls, I expect there is no file overwritten among `overlay`, `closed` and `jcl`, otherwise one of files can be eliminated completely;
* I feel it is better to add `INCLUDE-IF` for each java file within `$(TOPDIR)/closed/src`, there is a chance that an intended JPP flag is ignored silently like https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/494#discussion_r849675153

@keithc-ca pls have another look.

JasonFengJ9: Added `-includeIfUnsure -noWarnIncludeIf` options for `$(TOPDIR)/closed`.

JasonFengJ9: Removed extra `FixPath` calls.

closed/GensrcJ9JCL.gmk
```diff
@@ -78,24 +86,12 @@ $(J9JCL_SOURCES_DONEFILE) : $(AllJclSource) $(AllDdrSource) $(OPENJDK_STAGED_PRO
...
-		com.ibm.jpp.commandline.CommandlineBuilder \
-			-includeIfUnsure \
-			-noWarnIncludeIf \
-			-verdict \
-			-baseDir "$(call FixPath,$(OPENJ9_TOPDIR)/debugtools)/" \
-			-config DDR_VM \
-			-srcRoot DDR_VM/ \
-			-xml "$(call FixPath,$(OPENJ9_TOPDIR)/jcl/jpp_configuration.xml)/" \
-			-dest "$(call FixPath,$(J9JCL_SOURCES_DIR)/openj9.dtfj/share/classes)" \
-			-macro:define "JAVA_SPEC_VERSION=$(VERSION_FEATURE)" \
```
keithc-ca, line 97: The option `-macro:define "JAVA_SPEC_VERSION=$(VERSION_FEATURE)"` should be retained for DDR_VM.

JasonFengJ9: Added the option into `jpp_generate_sources` which is not harmful to other configurations already having such macro define.

tajila:
@keithc-ca Is there anything else left for this?

closed/GensrcJ9JCL.gmk
JasonFengJ9: Thanks @keithc-ca suggestions, this is ready for another look.

keithc-ca (approved):

keithc-ca:
jenkins compile alinux64 jdk11

PR #503, opened 2022-04-14
Invoke InternalCRIUSupport.isCRIUCapable() to avoid loadNativeLibrary()

JasonFengJ9:
Also improve `getenv(String name)` when `tracePrunedEnvVarsValue` is not set.

closes https://github.com/eclipse-openj9/openj9/issues/14789
depends on https://github.com/eclipse-openj9/openj9/pull/14917 

### DIFF
--- a/src/java.base/unix/classes/java/lang/ProcessEnvironment.java
+++ b/src/java.base/unix/classes/java/lang/ProcessEnvironment.java
@@ -73,8 +73,8 @@ final class ProcessEnvironment
     private static final Map<String,String> theUnmodifiableEnvironment;
     static final int MIN_NAME_LENGTH = 0;
 /*[IF CRIU_SUPPORT]*/
-    // CRIU enable flag
-    private static final boolean isCRIUEnabled;
+    // CRIU capable flag
+    private static final boolean isCRIUCapable;
     // 1 - prints a message if the env var was set but not in the immutable list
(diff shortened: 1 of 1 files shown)


src/java.base/unix/classes/java/lang/ProcessEnvironment.java
```diff
@@ -86,11 +86,11 @@
         // to putenv/setenv from C will not be visible from Java code.
         byte[][] environ = environ();
 /*[IF CRIU_SUPPORT]*/
-        isCRIUEnabled = InternalCRIUSupport.isCRIUSupportEnabled();
+        isCRIUCapable = InternalCRIUSupport.isCRIUCapable();
```
tajila, line 89: isnt `isCRIUCapable` always going to return true here?

JasonFengJ9: hmm, will have to rework on it, and verify via perf tests.
Marking this as draft for now.

JasonFengJ9: Updated [Invoke isCRIUSupportEnabledImpl() w/o loading j9criu native library](https://github.com/eclipse-openj9/openj9/pull/14917) which obsoletes this PR.

@tajila please have another look.

PR #519, opened 2022-05-05
AIX with xlc 16 must set -std=c99

pshipton:
Otherwise there is an error '-std=gnu++0x' not allowed with 'C/ObjC'

See https://github.com/eclipse-openj9/openj9/pull/15007

### DIFF
--- a/make/autoconf/flags-cflags.m4
+++ b/make/autoconf/flags-cflags.m4
@@ -23,6 +23,10 @@
 # questions.
 #
 
+# ===========================================================================
+# (c) Copyright IBM Corp. 2022, 2022 All Rights Reserved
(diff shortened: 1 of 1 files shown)


pshipton:
@keithc-ca any comments on this before I run more extensive testing?

keithc-ca (approved):
I might have expressed this differently, but it's probably better to retain as much of the structure as possible from newer JDK versions so as to reduce the merge cost if the related changed is back-ported upstream.

pshipton:
Removing from draft. This doesn't need to be merged yet, but doing so won't hurt anything as it's merging to the staging branch which already doesn't work by itself.

PR #520, opened 2022-05-09
FIPS Support for Linux x86 RHEL

taoliult:
Signed-off-by: Tao Liu <tao.liu@ibm.com>

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/FIPSConfigurator.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/FIPSConfigurator.java
@@ -1,0 +1,159 @@
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2022, 2022 All Rights Reserved
+ * ===========================================================================
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * IBM designates this particular file as subject to the "Classpath" exception
+ * as provided by IBM in the LICENSE file that accompanied this code.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
+ *
+ * ===========================================================================
+ */
+
+package openj9.internal.security;
+
+import java.util.Iterator;
+import java.util.Map.Entry;
+import java.util.Properties;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+
+import sun.security.util.Debug;
+
+/**
+ * Configures the security providers when in FIPS mode.
+ */
+public final class FIPSConfigurator {
+
+    private static final Debug debug = Debug.getInstance("semerufips");
+
+    // FIPS mode enable check, only supported on Linux x64.
+    private static final boolean userEnabledFIPS;
+    private static final boolean isFIPSSupported;
+    private static final boolean shouldEnableFIPS;
+
+    static {
+        String[] props = AccessController.doPrivileged(
+                new PrivilegedAction<>() {
+                    @Override
+                    public String[] run() {
+                        return new String[] {System.getProperty("semeru.fips"),
+                                System.getProperty("os.name"),
+                                System.getProperty("os.arch")};
+                    }
+                });
+        userEnabledFIPS = Boolean.parseBoolean(props[0]);
+        isFIPSSupported = "Linux".equalsIgnoreCase(props[1])
+                && "amd64".equalsIgnoreCase(props[2]);
+        shouldEnableFIPS = userEnabledFIPS && isFIPSSupported;
+    }
+
+    private FIPSConfigurator() {
+        super();
+    }
+
+    /**
+     * FIPS mode will be enabled only if the semeru.fips system
+     * property is true (default as false).
+     *
+     * @return true if FIPS is enabled
+     */
+    public static boolean enableFips() {
+        return shouldEnableFIPS;
+    }
+
+    /**
+     * Remove the security providers and only add the FIPS security providers.
+     *
+     * @param props the java.security properties
+     * @return true if the FIPS properties loaded successfully
+     */
+    public static boolean configureFIPS(Properties props) {
+        boolean loadedProps = false;
+
+        // Check if FIPS is supported on this platform.
+        if (userEnabledFIPS && !isFIPSSupported) {
+            throw new RuntimeException("FIPS is not supported on this platform.");
+        }
+
+        try {
+            if (shouldEnableFIPS) {
+                if (debug != null) {
+                    debug.println("FIPS mode detected, loading properties");
+                }
+
+                // Remove all security providers.
+                Iterator<Entry<Object, Object>> i = props.entrySet().iterator();
+                while (i.hasNext()) {
+                    Entry<Object, Object> e = i.next();
+                    if (((String) e.getKey()).startsWith("security.provider")) {
+                        if (debug != null) {
+                            debug.println("Removing provider: " + e);
+                        }
+                        i.remove();
+                    }
+                }
+
+                // Add FIPS security providers.
+                props.put("security.provider.1", "SunPKCS11 ${java.home}/conf/security/nss.fips.cfg");
+                props.put("security.provider.2", "SUN");
+                props.put("security.provider.3", "SunEC");
+                props.put("security.provider.4", "SunJSSE");
+
+                // Add FIPS security properties.
+                props.put("keystore.type", "PKCS11");
+                System.setProperty("javax.net.ssl.keyStore", "NONE");
+
+                // Add FIPS disabled algorithms.
+                String disabledAlgorithms = props.get("jdk.tls.disabledAlgorithms")
+                        + ", X25519, X448"
+                        + ", SSLv3, TLSv1, TLSv1.1"
+                        + ", TLS_CHACHA20_POLY1305_SHA256"
+                        + ", TLS_DHE_RSA_WITH_AES_256_GCM_SHA384, TLS_DHE_DSS_WITH_AES_256_GCM_SHA384"
+                        + ", TLS_DHE_RSA_WITH_AES_128_GCM_SHA256, TLS_DHE_DSS_WITH_AES_128_GCM_SHA256"
+                        + ", TLS_DHE_RSA_WITH_AES_256_CBC_SHA256, TLS_DHE_DSS_WITH_AES_256_CBC_SHA256"
+                        + ", TLS_DHE_RSA_WITH_AES_128_CBC_SHA256, TLS_DHE_DSS_WITH_AES_128_CBC_SHA256"
+                        + ", TLS_DHE_RSA_WITH_AES_256_CBC_SHA, TLS_DHE_DSS_WITH_AES_256_CBC_SHA"
+                        + ", TLS_DHE_RSA_WITH_AES_128_CBC_SHA, TLS_DHE_DSS_WITH_AES_128_CBC_SHA"
+                        + ", TLS_RSA_WITH_AES_256_GCM_SHA384, TLS_RSA_WITH_AES_128_GCM_SHA256"
+                        + ", TLS_RSA_WITH_AES_256_CBC_SHA256, TLS_RSA_WITH_AES_128_CBC_SHA256"
+                        + ", TLS_RSA_WITH_AES_256_CBC_SHA, TLS_RSA_WITH_AES_128_CBC_SHA"
+                        + ", TLS_AES_256_GCM_SHA384, TLS_AES_128_GCM_SHA256"
+                        + ", TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384, TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384"
+                        + ", TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256, TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256"
+                        + ", TLS_ECDH_ECDSA_WITH_AES_256_GCM_SHA384, TLS_ECDH_RSA_WITH_AES_256_GCM_SHA384"
+                        + ", TLS_ECDH_ECDSA_WITH_AES_128_GCM_SHA256, TLS_ECDH_RSA_WITH_AES_128_GCM_SHA256"
+                        + ", TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256"
+                        + ", TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256"
+                        + ", TLS_DHE_RSA_WITH_CHACHA20_POLY1305_SHA256"
+                        + ", TLS_EMPTY_RENEGOTIATION_INFO_SCSV";
+                props.put("jdk.tls.disabledAlgorithms", disabledAlgorithms);
+
+                if (debug != null) {
+                    debug.println("FIPS mode properties loaded");
+                    debug.println(props.toString());
+                }
+
+                loadedProps = true;
+            }
+        } catch (Exception e) {
+            if (debug != null) {
+                debug.println("Unable to load FIPS configuration");
+                e.printStackTrace();
+            }
+        }
+        return loadedProps;
+    }
+}
--- a/src/java.base/share/classes/java/security/SecureRandom.java
+++ b/src/java.base/share/classes/java/security/SecureRandom.java
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2022, 2022 All Rights Reserved
+ * ===========================================================================
+ */
+
 package java.security;
 
 import java.util.*;
@@ -34,6 +40,8 @@ import sun.security.jca.*;
 import sun.security.jca.GetInstance.Instance;
 import sun.security.provider.SunEntries;
 import sun.security.util.Debug;
+
+import openj9.internal.security.FIPSConfigurator;
 
 /**
  * This class provides a cryptographically strong random number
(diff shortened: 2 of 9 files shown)


alon-sh:
this PR enables FIPS support to IBM Semeru for RHEL x86 linux only, it requires putting RHEL into FIPS mode first and then instructing IBM Semeru via the command line `semeru.fips=true`.

The automated testing for this mode will come online by end of the month fyi @llxia

JDK 17/18 PRs will be created once review for JDK11 is done.

fyi @mstoodle

src/java.base/share/classes/java/security/FIPSConfigurator.java
```diff
@@ -0,0 +1,152 @@
...
+import java.util.List;
+
+import sun.security.util.Debug;
+
+/**
+ * Configures the security providers when in FIPS mode.
+ */
+public final class FIPSConfigurator {
+
+    static final Debug debug = Debug.getInstance("semerufips");
```
keithc-ca, line 39: This field should be `private`.

src/java.base/share/classes/java/security/FIPSConfigurator.java
```diff
@@ -0,0 +1,152 @@
...
+
+    // FIPS mode enable check, only support on Linux x64
+    private static boolean userEnabledFIPS = false;
+    private static boolean isFIPSSupported = false;
+    private static boolean shouldEnableFIPS = false;
```
keithc-ca, line 44: These should be `final` and initialized only in the static initializer.

src/java.base/share/classes/java/security/FIPSConfigurator.java
```diff
@@ -0,0 +1,152 @@
...
+                new PrivilegedAction<>() {
+                    @Override
+                    public List<String> run() {
+                        return List.of(
+                                System.getProperty("semeru.fips", "false"),
+                                System.getProperty("os.name"),
+                                System.getProperty("os.arch"));
+                    }
+                });
+        userEnabledFIPS = Boolean.valueOf(props.get(0));
```
keithc-ca, line 57: This should use `parseBoolean()` rather than `valueOf()`.

src/java.base/share/classes/java/security/FIPSConfigurator.java
```diff
@@ -0,0 +1,152 @@
...
+        isFIPSSupported = "Linux".equalsIgnoreCase(props.get(1))
+                && "amd64".equalsIgnoreCase(props.get(2));
+        shouldEnableFIPS = userEnabledFIPS && isFIPSSupported;
+    }
+
+    /**
+     * FIPS mode will be enabled only if the semeru.fips system
+     * property is true (default as false)
```
keithc-ca, line 65: Please finish each sentence with a period.

src/java.base/share/classes/java/security/FIPSConfigurator.java
```diff
@@ -0,0 +1,152 @@
...
+                props.put("security.provider.4", "SunJSSE");
+
+                // Add FIPS security properties
+                props.put("keystore.type", "PKCS11");
+                System.setProperty("javax.net.ssl.keyStore", "NONE");
+
+                // Add FIPS disabled algorithms
+                StringBuffer disabledAlgorithmsListFIPS = new StringBuffer();
+                String disabledAlgorithmsList = (String) props.get("jdk.tls.disabledAlgorithms");
+                disabledAlgorithmsListFIPS.append(disabledAlgorithmsList);
```
keithc-ca, line 113: Prefer `StringBuilder` over `StringBuffer`, but neither are required here:
```
    String disabledAlgorithms = props.get("jdk.tls.disabledAlgorithms")
            + ", X25519, X448"
            + ", SSLv3, TLSv1, TLSv1.1"
            + ", TLS_CHACHA20_POLY1305_SHA256, "
            + "TLS_DHE_RSA_WITH_AES_256_GCM_SHA384, TLS_DHE_DSS_WITH_AES_256_GCM_SHA384, "
            + "TLS_DHE_RSA_WITH_AES_128_GCM_SHA256, TLS_DHE_DSS_WITH_AES_128_GCM_SHA256, "
            + "TLS_DHE_RSA_WITH_AES_256_CBC_SHA256, TLS_DHE_DSS_WITH_AES_256_CBC_SHA256, "
            + "TLS_DHE_RSA_WITH_AES_128_CBC_SHA256, TLS_DHE_DSS_WITH_AES_128_CBC_SHA256, "
            + "TLS_DHE_RSA_WITH_AES_256_CBC_SHA, TLS_DHE_DSS_WITH_AES_256_CBC_SHA, "
            + "TLS_DHE_RSA_WITH_AES_128_CBC_SHA, TLS_DHE_DSS_WITH_AES_128_CBC_SHA, "
            + "TLS_RSA_WITH_AES_256_GCM_SHA384, TLS_RSA_WITH_AES_128_GCM_SHA256, "
            + "TLS_RSA_WITH_AES_256_CBC_SHA256, TLS_RSA_WITH_AES_128_CBC_SHA256, "
            + "TLS_RSA_WITH_AES_256_CBC_SHA, TLS_RSA_WITH_AES_128_CBC_SHA, "
            + "TLS_AES_256_GCM_SHA384, TLS_AES_128_GCM_SHA256, "
            + "TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384, TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384, "
            + "TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256, TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256, "
            + "TLS_ECDH_ECDSA_WITH_AES_256_GCM_SHA384, TLS_ECDH_RSA_WITH_AES_256_GCM_SHA384, "
            + "TLS_ECDH_ECDSA_WITH_AES_128_GCM_SHA256, TLS_ECDH_RSA_WITH_AES_128_GCM_SHA256, "
            + "TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256, "
            + "TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256, "
            + "TLS_DHE_RSA_WITH_CHACHA20_POLY1305_SHA256, "
            + "TLS_EMPTY_RENEGOTIATION_INFO_SCSV";
    props.put("jdk.tls.disabledAlgorithms", disabledAlgorithms);
```

src/java.base/share/classes/java/security/FIPSConfigurator.java
```diff
@@ -0,0 +1,152 @@
...
+            }
+        } catch (Exception e) {
+            if (debug != null) {
+                debug.println("Unable to load FIPS configuration");
+                e.printStackTrace();
+            }
+        }
+        return loadedProps;
+    }
+}
```
keithc-ca, line 152: Each non-empty file should end with a new-line.

src/java.base/share/classes/java/security/SecureRandom.java
```diff
@@ -154,6 +160,8 @@ public class SecureRandom extends java.util.Random {
     private static final boolean skipDebug =
         Debug.isOn("engine=") && !Debug.isOn("securerandom");
 
+    private static boolean isFips = FIPSConfigurator.enableFips();
```
keithc-ca, line 163: Please remove this - it's only used once - use `FIPSConfigurator.enableFips()` directly instead.

src/java.base/share/classes/sun/security/provider/SunEntries.java
```diff
@@ -34,6 +34,7 @@
 import java.net.*;
 import java.util.*;
 import java.security.*;
+import java.security.FIPSConfigurator;
```
keithc-ca, line 37: This import is redundant.

src/java.base/share/classes/sun/security/provider/SunEntries.java
```diff
@@ -95,6 +96,8 @@ public final class SunEntries {
     // the default algo used by SecureRandom class for new SecureRandom() calls
     public static final String DEF_SECURE_RANDOM_ALGO;
 
+    private static boolean isFips = FIPSConfigurator.enableFips();
```
keithc-ca, line 99: Please remove this in favour of using `FIPSConfigurator.enableFips()`.

src/java.base/share/classes/sun/security/provider/SunEntries.java
```diff
@@ -115,218 +118,266 @@ public static List<String> createAliasesWithOid(String ... oids) {
...
-        if (NativePRNG.isAvailable()) {
-            add(p, "SecureRandom", "NativePRNG",
-                    "sun.security.provider.NativePRNG",
+        if (isFips) {
+            // FIPS supported
+            attrs.put("ImplementedIn", "Software");
+
+            /*
+            * Certificates
+            */
```
keithc-ca, line 127: Please fix indentation of all comments (the asterisks should line up):
```
            /*
             * Certificates
             */
```

src/java.base/share/classes/sun/security/provider/SunEntries.java
```diff
@@ -115,218 +118,266 @@ public static List<String> createAliasesWithOid(String ... oids) {
...
+                    "sun.security.provider.DSA$SHA1withDSA",
+                    createAliasesWithOid("1.2.840.10040.4.3", "DSA", "DSS",
+                        "SHA/DSA", "SHA-1/DSA", "SHA1/DSA", "SHAwithDSA",
+                        "DSAWithSHA1", "1.3.14.3.2.13", "1.3.14.3.2.27"), attrs);
+            add(p, "Signature", "NONEwithDSA", "sun.security.provider.DSA$RawDSA",
+                    createAliases("RawDSA"), attrs);
+
+            attrs.put("KeySize", "2048"); // for SHA224 and SHA256 DSA signatures
+
+            add(p, "Signature", "SHA1withDSA",
```
keithc-ca, line 212: Please explain why this should change for non-FIPS mode.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
```diff
@@ -53,6 +59,8 @@ public final class SunEC extends Provider {
 
     private static final long serialVersionUID = -2279741672933606418L;
 
+    private static boolean isFips = FIPSConfigurator.enableFips();
```
keithc-ca, line 62: Please remove this and use `FIPSConfigurator.enableFips()` directly.

keithc-ca (changes requested):

keithc-ca:
There could be more detail in the commit message, but in any case there should be a blank line after the summary and before anything else. Even if you add nothing else it would be:
```
Add FIPSConfigurator and FIPS algorithms check

Signed-off-by: Tao Liu <tao.liu@ibm.com>
```

src/java.base/share/classes/java/security/FIPSConfigurator.java
keithc-ca: Please complete the javadoc to avoid this warning:
```
support/j9jcl/java.base/share/classes/java/security/FIPSConfigurator.java:66: warning: no @return
    public static boolean enableFips() {
                          ^
```

make/copy/Copy-java.base.gmk
```diff
@@ -253,3 +257,20 @@ $(eval $(call SetupTextFileProcessing, CREATE_CLASSFILE_CONSTANTS_H, \
...
+  NSS_FIPS_CFG_SRC_LIST := $(NSS_FIPS_CFG_SRC)
+
+  $(NSS_FIPS_CFG_DST): $(NSS_FIPS_CFG_SRC_LIST)
+		$(MKDIR) -p $(@D)
+		$(RM) $@ $@.tmp
+		$(foreach f,$(NSS_FIPS_CFG_SRC_LIST),$(CAT) $(f) >> $@.tmp;)
+		$(MV) $@.tmp $@
+
+  TARGETS += $(NSS_FIPS_CFG_DST)
+endif
```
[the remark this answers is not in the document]
taoliult: Updated

keithc-ca (changes requested):
Please also address the line-end check failure.

src/java.base/share/classes/sun/security/provider/SunEntries.java
keithc-ca: Can you please answer this query?

closed/src/java.base/share/classes/java/security/FIPSConfigurator.java
```diff
@@ -0,0 +1,156 @@
...
+    private static final boolean userEnabledFIPS;
+    private static final boolean isFIPSSupported;
+    private static final boolean shouldEnableFIPS;
+
+    static {
+        List<String> props = AccessController.doPrivileged(
+                new PrivilegedAction<>() {
+                    @Override
+                    public List<String> run() {
+                        return Arrays.asList(
```
keithc-ca, line 52: Rather than `List.of()` or `Arrays.asList()`, why not just return `String[]`?

closed/src/java.base/share/classes/java/security/FIPSConfigurator.java
```diff
@@ -0,0 +1,154 @@
...
+    private static final Debug debug = Debug.getInstance("semerufips");
+
+    // FIPS mode enable check, only supported on Linux x64.
+    private static final boolean userEnabledFIPS;
+    private static final boolean isFIPSSupported;
+    private static final boolean shouldEnableFIPS;
+
+    static {
+        String[] props = AccessController.doPrivileged(
+                new PrivilegedAction<String[]>() {
```
keithc-ca, line 47: The type parameter (`String[]`) is not needed here.

closed/src/java.base/share/classes/java/security/FIPSConfigurator.java
```diff
@@ -0,0 +1,154 @@
...
+                    @Override
+                    public String[] run() {
+                        String[] arr = {System.getProperty("semeru.fips"),
+                                System.getProperty("os.name"),
+                                System.getProperty("os.arch")};
+                        return arr;
```
keithc-ca, line 53: Please inline the array expression (and remove `arr`).

keithc-ca (changes requested):
I'm still awaiting an answer to my question here: https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/520#discussion_r868499457.

keithc-ca:
@taoliult Could you please answer my question: https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/520#discussion_r868499457?

src/java.base/share/classes/sun/security/provider/SunEntries.java
alon-sh: That specific code disables non-certified FIPS cryptographic algorithms in the SUN security provider once the JDK is placed in FIPS mode. 

The reason we don't disable the entire security provider is that there are some needed services in the SUN provider that are not cryptographic in nature and are needed - for example, the security policy service provider and some certificate manipulation services.

keithc-ca: This code applies only if `FIPSConfigurator.enableFips()` answers `false`, so the JDK is not in FIPS mode, or do I misunderstand something?

alon-sh: the code is 

```
if (isFips) {
.. add only specific services that are needed in FIPS mode ...
} else {
.. add everything that SUN provider supports, including non FIPS mode ....
}
```

so when in isFIPS==true, only those specific services are activated and everything else is not.

alon-sh: i dont believe it is changed - just the diff got messed up

keithc-ca: It did change, from
```
        add(p, "Signature", "SHA224withDSA",
```
to
```
            add(p, "Signature", "SHA1withDSA",
```
The indentation change I expected, but not the change from `"SHA224withDSA"` to `"SHA1withDSA"`.
Looking at https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/520/files?w=1 should clarify that.

alon-sh: @taoliult please fix

alon-sh:
we can make it a private class @taoliult it doesnt need to be accessible to the outside

keithc-ca:
It can't be package-private: it's used by classes in two other packages:
```
sun.security.provider.SunEntries
sun.security.ec.SunEC
```
If need be, `FIPSConfigurator` can be moved to an internal package and exported to `jdk.crypto.ec`.

pshipton:
It can't be public in java.security. It can be accessed by other packages via some internal API.

pshipton:
> `FIPSConfigurator` can be moved to an internal package and exported to `jdk.crypto.ec`

Yes, this sounds good.

alon-sh:
@pshipton @keithc-ca could you point us at similar code that we can use as an example for this change?

keithc-ca:
I suggest moving `FIPSConfigurator` to `openj9.internal.security` and references updated accordingly.
The clause `exports openj9.internal.security to jdk.crypto.ec;` would be added to [java.base/module-info](https://github.com/eclipse-openj9/openj9/blob/master/jcl/src/java.base/share/classes/module-info.java.extra).

pshipton:
If FIPSConfigurator is moved to OpenJ9, certain things needs to change, like `semerufips` and `semeru.fips`. It's not appropriate to reference Semeru from OpenJ9. Semeru is IBM's build of OpenJ9 + OpenJDK, others could build it as well and name it something different. I think it should just stay on the openjdk side considering all the security provider config in there.

pshipton:
Really this PR shouldn't refer to Semeru either, for the same reasons.

keithc-ca:
I wasn't suggesting it move to the openj9 repository, just suggesting that the package name should change.

pshipton:
Then https://github.com/eclipse-openj9/openj9/blob/master/jcl/src/java.base/share/classes/module-info.java.extra is the wrong place to add the export.

taoliult:
Changed the package of FIPSConfigurator to openj9.internal.security. And the method configureFIPS changed to public, since Security.java and FIPSConfigurator.java are not in the same package.

pshipton:
What happens if you add the command line parameter `--add-exports java.base/openj9.internal.security=jdk.crypto.ec`

Was it a full build or an incremental build? You can check the exports by running `javap --module java.base module-info` to confirm the export took effect.

pshipton:
The permission needs to be added to https://github.ibm.com/runtimes/openj9-openjdk-jdk11/blob/ibm_sdk/src/java.base/share/lib/security/default.policy#L131

closed/src/java.base/share/classes/openj9/internal/security/FIPSConfigurator.java
```diff
@@ -0,0 +1,154 @@
...
+ *
+ * ===========================================================================
+ */
+
+package openj9.internal.security;
+
+import java.util.Iterator;
+import java.util.Map.Entry;
+import java.util.Properties;
+import java.security.*;
```
keithc-ca, line 30: Please don't use wildcard imports: import the two needed classes explicitly instead.

closed/src/java.base/share/classes/openj9/internal/security/FIPSConfigurator.java
```diff
@@ -0,0 +1,154 @@
...
+
+import sun.security.util.Debug;
+
+/**
+ * Configures the security providers when in FIPS mode.
+ */
+public final class FIPSConfigurator {
```
keithc-ca, line 37: This class should have an explicit constructor; it can be `private` if it doesn't need to be instantiated.

src/java.base/share/classes/module-info.java
```diff
@@ -22,6 +22,11 @@
  * or visit www.oracle.com if you need additional information or have any
  * questions.
  */
+/*
```
keithc-ca, line 25: Please leave a blank line before the copyright notice (like you added in `SunEntries.java`).

keithc-ca: Please do this for each IBM copyright notice.

src/java.base/share/lib/security/default.policy
```diff
@@ -144,6 +146,7 @@ grant codeBase "jrt:/jdk.crypto.cryptoki" {
                    "accessClassInPackage.sun.security.*";
     permission java.lang.RuntimePermission "accessClassInPackage.sun.nio.ch";
     permission java.lang.RuntimePermission "loadLibrary.j2pkcs11";
+    permission java.lang.RuntimePermission "getClassLoader";
```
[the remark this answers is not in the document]
keithc-ca: For the record, this is related to eclipse-openj9/openj9#15078 which will be addressed separately.

keithc-ca:
jenkins test sanity,extended osx,xlinux jdk11

keithc-ca (approved):

PR #545, opened 2022-06-28
Default to using cmake on cross-compiles

dnakamura:


closed/autoconf/custom-hook.m4
```diff
@@ -62,18 +62,7 @@ AC_DEFUN([OPENJ9_CONFIGURE_CMAKE],
...
-            with_cmake=cmake
-          else
-            with_cmake=no
-          fi
-          ;;
-        *)
-          with_cmake=no
-          ;;
-      esac
+      with_cmake=cmake
```
pshipton, line 65: This should be simplified more than it is. As it stands, we are testing if `--with-cmake` is yes or not set, and then setting cmake, but also setting cmake in the else case.

pshipton: Or to put it another way, this change breaks `--with-cmake=no`

pshipton:
The title of this commit is not appropriate. The intention is not to make cmake the default for cross-compiles, but for all builds.

closed/autoconf/custom-hook.m4
keithc-ca: No, I don't think so. The third and fourth arguments to `AC_ARG_WITH` are for when `--with-cmake` is used and for when it is not, respectively. `--with-cmake=no` (and `--without-cmake`) will be handled by the third argument which will leave `with_cmake=no` alone.

keithc-ca (approved):

keithc-ca:
jenkins compile osx jdk11

knn-k:
CMake does not work with AArch64 Linux cross builds.  See the following errors:

With DDR enabled:
```
CMake Error at /root/openj9/openj9-openjdk-jdk11/omr/cmake/modules/OmrDDRSupport.cmake:36 (message):
  DDR is not supported when cross-compiling.  Use -DOMR_DDR=OFF to disable
  DDR.
```

With DDR disabled:
```
Scanning dependencies of target hookgen
[  3%] Building CXX object runtime/omr/tools/hookgen/CMakeFiles/hookgen.dir/HookGen.cpp.o
[  3%] Building CXX object runtime/omr/tools/hookgen/CMakeFiles/hookgen.dir/main.cpp.o
[  3%] Linking CXX executable ../../../hookgen
[  3%] Built target hookgen
[  3%] Generating ut_j9thr.c, ut_j9thr.h, ut_j9thr.pdat
/lib/ld-linux-aarch64.so.1: No such file or directory
```
hookgen needs to be built statically for running with qemu.

knn-k:
I found I was able to build for AArch64 Linux with CMake in a cross-building environment.
See https://github.com/eclipse-openj9/openj9/issues/10266#issuecomment-1170621163

PR #559, opened 2022-07-25
Fix java/util/zip/DeflateIn_InflateOut.skipBytes Test

r30shah:
java/util/zip/DeflateIn_InflateOut.skipBytes verifies functionality of
DeflaterInputStream.skipBytes. While testing skipping of bytes, test
uses different input data with same data length, and compares the
compressed sizes of these data to verify skipping bytes. As per zlib
specification which is what is used by Deflater in java, length of
compressed data is highly dependable on input data contents, so it will
be incorrect to assume that compression ratio would be constant for same
length input data. For more details please see explaination in
https://github.com/eclipse-openj9/openj9/issues/14948#issuecomment-1194616794.

Signed-off-by: Rahil Shah <rahil@ca.ibm.com>

### DIFF
--- a/test/jdk/java/util/zip/DeflateIn_InflateOut.java
+++ b/test/jdk/java/util/zip/DeflateIn_InflateOut.java
@@ -41,9 +41,7 @@ public class DeflateIn_InflateOut {
     private static ByteArrayOutputStream baos;
     private static InflaterOutputStream ios;
 
-    private static void reset() {
-        new Random(new Date().getTime()).nextBytes(data);
-
+    private static void resetStreams() {
         bais = new ByteArrayInputStream(data);
         dis = new DeflaterInputStream(bais);
 
@@ -51,11 +49,16 @@ public class DeflateIn_InflateOut {
         ios = new InflaterOutputStream(baos);
     }
 
+    private static void resetAll() {
+        new Random(new Date().getTime()).nextBytes(data);
+        resetStreams();
+    }
+
     /** Check byte arrays read/write. */
     private static void ArrayReadWrite() throws Throwable {
         byte[] buf = new byte[512];
 
-        reset();
+        resetAll();
         check(dis.available() == 1);
         for (;;) {
             int len = dis.read(buf, 0, buf.length);
@@ -74,7 +77,7 @@ public class DeflateIn_InflateOut {
     private static void ArrayReadByteWrite() throws Throwable {
         byte[] buf = new byte[512];
 
-        reset();
+        resetAll();
         for (;;) {
             int len = dis.read(buf, 0, buf.length);
             if (len <= 0) {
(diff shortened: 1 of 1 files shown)


r30shah:
@keithc-ca / @pshipton  I have marked this as WIP, till we test it through grinder and figure out delivery process (First in OpenJDK master or external). Would appreciate your review to polish it off.

pshipton:
I don't have any concerns accepting a test change before submitting it to OpenJDK. There should be comments included in the code that explain the changes.

r30shah:
@pshipton I have added a  comment in the code to mention why input data is fixed for checking skipping of different number of bytes in `skipBytes`. Can you review the change. 

Also wanted to check with @backwaterred if it is possible to contribute this to OpenJDK repo?

test/jdk/java/util/zip/DeflateIn_InflateOut.java
```diff
@@ -149,7 +152,9 @@ private static void SkipBytes() throws Throwable {
         int numReadable = 0;
 
         // Count number of bytes that are read
-        reset();
+
+		// Reset Input data and Streams
```
pshipton, line 156: Not sure why this one got a comment when the others that changed from reset() to resetAllI() didn't. If this comment is retained, spaces should be used for indenting rather than tabs.

pshipton:
@keithc-ca I recall you reviewed a similar change elsewhere, if you want to look at this one.

test/jdk/java/util/zip/DeflateIn_InflateOut.java
pshipton: I'm fine if this comment is just removed.

keithc-ca (approved):

test/jdk/java/util/zip/DeflateIn_InflateOut.java
r30shah: Removed comment in https://github.com/ibmruntimes/openj9-openjdk-jdk11/compare/6d92003cfe8e2e8d76535edcd9d18fcd9aa4b70c..ee1497e7a2ee0a56a59c89136d15b5adf9d41c73

pshipton:
Pls port it to jdk17, 19, next.

pshipton:
also 8

pshipton (approved):

r30shah:
@pshipton  Looking at the source code for same test in JDK17,JDK19 and JDKNext [1][2][3], seems like they have already fixed this.  Though it keeps the data constant (One of the suggestion @keithc-ca  had was to perform different test with different sets of data. I did port the change back in https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/609 though. Let me know what do you think. Use the fix in this PR for JDK8 and JDK11 which would perform test with different input data (Only keep data consistent for skipBytes test which was failing) or want to port the changes from JDKnext. 

[1]. https://github.com/r30shah/openj9-openjdk-jdk17/commit/66cd6d5f673f211ee8c088b68b036b5cff7b4d12

keithc-ca:
Perhaps we should be applying fixes closer to the change made in jdk17?

pshipton:
Backporting the changes from jdk17+ is preferred.

Issue #547, opened 2022-07-11
Update the OpenJDK exclude tests list for FIPS after jdk_security3 test target enabled


taoliult:
The FIPS mode has restrictive security providers, so update the OpenJDK exclude tests list “ProblemList-fips.txt” for the FIPS testing, after jdk_security3 test target enabled.

taoliult:
For the extended.openjdk exclude list:

- NoSuchAlgorithmException: no such algorithm: DSA for provider SUN
- NoSuchAlgorithmException: no such algorithm: MD2 for provider SUN
- NoSuchAlgorithmException: no such algorithm: SHA for provider SUN
- NoSuchAlgorithmException: no such algorithm: SHA-256 for provider SUN
- NoSuchAlgorithmException: no such algorithm: MD5 for provider SUN
- XML related

sun/security/pkcs11/Signature/TestDSAKeyLength.java.TestDSAKeyLength
sun/security/pkcs11/MessageDigest/ReinitDigest.java.ReinitDigest
sun/security/provider/NSASuiteB/TestSHAwithDSASignatureOids.java.TestSHAwithDSASignatureOids
sun/security/provider/NSASuiteB/TestSHAOids.java.TestSHAOids
sun/security/provider/NSASuiteB/TestDSAGenParameterSpecLongKey.java.TestDSAGenParameterSpecLongKey
sun/security/provider/NSASuiteB/TestDSAGenParameterSpec.java.TestDSAGenParameterSpec
sun/security/provider/MessageDigest/TestSHAClone.java.TestSHAClone
sun/security/provider/MessageDigest/Offsets.java.Offsets
sun/security/provider/MessageDigest/DigestKAT.java.DigestKAT
sun/security/provider/DSA/TestLegacyDSAKeyPairGenerator.java.TestLegacyDSAKeyPairGenerator
sun/security/provider/DSA/TestKeyPairGenerator.java.TestKeyPairGenerator
sun/security/provider/DSA/TestDSA2.java.TestDSA2
sun/security/provider/DSA/TestDSA.java.TestDSA
sun/security/provider/DSA/TestAlgParameterGenerator.java.TestAlgParameterGenerator
sun/security/provider/DSA/SupportedDSAParamGenLongKey.java.SupportedDSAParamGenLongKey
sun/security/provider/DSA/SupportedDSAParamGen.java.SupportedDSAParamGen
[...]

PR #560, opened 2022-07-26
Update OpenJDK11 ProblemList-fips for FIPS

taoliult:
When the jdk_security3 test target enabled, there are more test failures due to the tests are not for the FIPS mode. For example, some tests are using provider SunJCE which is removed from FIPS mode. Adding those tests into the exclude list so that those tests will not be run in the FIPS mode testing.

Signed-off-by: Tao Liu <tao.liu@ca.ibm.com>

### DIFF
--- a/test/jdk/ProblemList-fips.txt
+++ b/test/jdk/ProblemList-fips.txt
@@ -201,6 +201,7 @@ com/sun/crypto/provider/KeyFactory/TestProviderLeak.java	https://github.com/ibmr
 com/sun/crypto/provider/KeyFactory/PBKDF2HmacSHA1FactoryTest.java	https://github.com/ibmruntimes/openj9-openjdk-jdk11/issues/525	linux-x64
 com/sun/crypto/provider/KeyAgreement/UnsupportedDHKeys.java	https://github.com/ibmruntimes/openj9-openjdk-jdk11/issues/525	linux-x64
 com/sun/crypto/provider/KeyAgreement/TestExponentSize.java	https://github.com/ibmruntimes/openj9-openjdk-jdk11/issues/525	linux-x64
+com/sun/crypto/provider/KeyAgreement/SupportedDHParamGensLongKey.java	https://github.com/ibmruntimes/openj9-openjdk-jdk11/issues/547	linux-x64
 com/sun/crypto/provider/KeyAgreement/SupportedDHParamGens.java	https://github.com/ibmruntimes/openj9-openjdk-jdk11/issues/525	linux-x64
 com/sun/crypto/provider/KeyAgreement/SupportedDHKeys.java	https://github.com/ibmruntimes/openj9-openjdk-jdk11/issues/525	linux-x64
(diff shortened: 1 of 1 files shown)


taoliult:
Extended.openjdk Grinder test result before the problem list:
https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/26160/

Extended.openjdk Grinder test result after the problem list:
https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/26161/

alon-sh:
please explain why we dont want to fix this. if we indeed dont fix, modify comment to explain why more clearly.

# java.lang.RuntimeException: Unexpected provider: SunPKCS11-NSS-FIPS.

sun/security/ssl/HandshakeHash/HandshakeHashCloneExhaustion.java	https://github.com/ibmruntimes/openj9-openjdk-jdk11/issues/547	linux-x64

taoliult:
@alon-sh

In the test code HandshakeHashCloneExhaustion.java, it first add the specified provider by

```
Security.insertProviderAt(new MyProvider(), 1);
```

And then get the MessageDigest without specify any provider. Code as below. 

```
MessageDigest md = MessageDigest.getInstance(s);
String p = md.getProvider().getName();
if (!p.equals("MyProvider")) {
    throw new RuntimeException("Unexpected provider: " + p);
}
```

And after that, it compares if the MessageDigest provider is MyProvider. It is not, because in FIPS, the MessageDigest come from SunPKCS11. 

So, the test case throws "java.lang.RuntimeException: Unexpected provider: SunPKCS11-NSS-FIPS".

I will modify the comment about this test case, to have more clearly details.

taoliult:
@alon-sh  Comment updated with more clearly details. Please help to review.

alon-sh:
i dont understand why the FIPS mode make it fail. inserting a provider dynamically should still work in FIPS mode.

taoliult:
@alon-sh About this inserting MyProvider dynamically issue, I need some time to take a look at. So, for this PR, I removed this HandshakeHashCloneExhaustion.java from the exclude list, and add this back into my backlog, and I will take a look at this.

alon-sh:
ok fine, please have a committer merge it.

taoliult:
@alon-sh Thanks for the review.

@pshipton This PR is for updating the FIPS exclude tests list after the jdk_security3 test target enabled. Alon already reviewed it. Please help to merge the PR. Following are the Grinder test results after the code updates according to the review.

FIPS extended.openjdk Grinder test result before the problem list update:
https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/26239/

FIPS extended.openjdk Grinder test result after the problem list update:
https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/26240/

pshipton (approved):

pshipton:
There are still 66 failures after the update (917 before).

alon-sh:
@taoliult FYI please look at failures and update exclusion list

taoliult:
@pshipton @alon-sh 
Yes, the test result still show 66 failures after the exclude list update. But actually there are 33 failures, each failed case will rerun, so the number of result failures will double the actual failures. For these 33 failures, I am working on them one by one to confirm.

PR #561, opened 2022-07-27
Add NULL check for CRIU field retrieval

JasonFengJ9:
The `task` could be null.

An internal test expects `IllegalArgumentException` when `task` is `null` and `time < 0` [1].

[1] https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/20d13352ee5c9569416880e340d5c589161082f9/src/java.base/share/classes/java/util/Timer.java#L415-L417

### DIFF
--- a/src/java.base/share/classes/java/util/Timer.java
+++ b/src/java.base/share/classes/java/util/Timer.java
@@ -203,7 +203,9 @@ public class Timer {
             throw new IllegalArgumentException("Negative delay.");
         /*[IF CRIU_SUPPORT]*/
         // only tasks scheduled before Checkpoint to be adjusted
-        if (InternalCRIUSupport.getCheckpointRestoreNanoTimeDelta() == 0) {
+        if ((InternalCRIUSupport.getCheckpointRestoreNanoTimeDelta() == 0)
+            && (task != null)
+        ) {
             task.criuAdjustRequired = true;
         }
         /*[ENDIF] CRIU_SUPPORT*/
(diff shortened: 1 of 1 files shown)


tajila:
Do we have a test for this

JasonFengJ9:
There are [internal tests](https://jazz103.hursley.ibm.com:9443/jazz/web/projects/JTC-JAT#action=com.ibm.team.workitem.viewWorkItem&id=147799) covering two cases:
1. `task = null` - NPE is expected
2. `task = null` && `time < 0` - `IllegalArgumentException` is expected

tajila:
I meant a criu test

tajila:
Also, if `task = null - NPE is expected` do we need to do anything? the `task.criuAdjustRequired = true;` will thrown the NPE

JasonFengJ9:
CRIU test - https://github.com/eclipse-openj9/openj9/blob/master/test/functional/cmdLineTests/criu/src/org/openj9/criu/TimeChangeTest.java

> Also, if task = null - NPE is expected do we need to do anything? the task.criuAdjustRequired = true; will thrown the NPE

There is no problem for NPE test, just for test reference that the `task` is `null`.
This fix is for `IllegalArgumentException` when `time < 0`.

tajila:
isnt  the `(delay < 0)` check always done first?

JasonFengJ9:
Yes, the test in question escaped first check, and passed `10000 + Long.MAX_VALUE - System.currentTimeMillis()` which turns to negative after `System.currentTimeMillis()+delay`.

JasonFengJ9:
As per [javadoc - java/util/Timer.html#schedule(java.util.TimerTask,long)](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Timer.html#schedule(java.util.TimerTask,long))
```
IllegalArgumentException - if delay is negative, or delay + System.currentTimeMillis() is negative.
```
The later checking `time < 0` is required for the actual `time` within `sched(TimerTask task, long time, long period)`, however `task.criuAdjustRequired` can't be moved there due to other fixed date `schedule` methods.

tajila:
Okay, I understand now. Thanks

tajila:
Jenkins test sanity xlinuxcriu jdk17

tajila:
is there a jdk17 and jdknext version of this?

JasonFengJ9:
Created https://github.com/ibmruntimes/openj9-openjdk-jdk17/pull/129 & https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/477, please review.

PR #571, opened 2022-08-29
Set OpenJDK test property vm.musl to false in OpenJ9PropsExt

pshipton:
It's now used by java/lang/ProcessBuilder/Basic.java
https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9-staging/test/jdk/java/lang/ProcessBuilder/Basic.java#L40

The property is already set for jdk17+


keithc-ca:
For the record, references were added in cab31b3a9cd9d0957d9c90493408006a8232559d.

PR #577, opened 2022-09-21
Revert fix for skipBytes and backport fix from jdk17

r30shah:
First commit in this PR reverts the fix that was merged in https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/559 to fix the failures seen in test DeflateIn_InflateOut.skipBytes. The test was fixed in JDK17 and it is preferable to port fixes that were made in jdk17+ repository so second commit in this PR backports fix from JDK17 repository.


pshipton:
The test [passed](https://hyc-runtimes-jenkins.swg-devops.com/view/Test_grinder/job/Grinder/27905) on rhel9s390xrt-1. lgtm

pshipton (approved):

PR #586, opened 2022-10-18
Java Restricted Security Mode

taoliult:
Signed-off-by: Tao Liu tao.liu@ibm.com

This PR is for adding the codes of Java Restricted Security Mode.

The Java Restricted Security Mode is used to restrict the providers and algorithms, by the properties configured in the java.security file.

The properties will be of the following format:

```
RestrictedSecurity(n).desc.name = This string would be presented to the user when the audit setting is turned on, used to identify the policy in a helpful manner
RestrictedSecurity(n).desc.number = Security policy number
RestrictedSecurity(n).desc.policy = Policy details link
RestrictedSecurity(n).desc.sunsetDate = Date by which the policy will sunset

RestrictedSecurity(n).tls.disabledNamedCurves = Map to jdk.disabledNamedCurves
RestrictedSecurity(n).tls.disabledAlgorithms = Map to jdk.tls.disabledAlgorithms
RestrictedSecurity(n).tls.ephemeralDHKeySize = Map to jdk.tls.ephemeralDHKeySize
RestrictedSecurity(n).tls.legacyAlgorithms = Map to jdk.tls.legacyAlgorithms

RestrictedSecurity(n).jce.certpath.disabledAlgorithms = Map to jdk.certpath.disabledAlgorithms
RestrictedSecurity(n).jce.legacyAlgorithms = Map to jdk.security.legacyAlgorithm
RestrictedSecurity(n).jce.provider.1 = Map to security.provider
RestrictedSecurity(n).jce.provider.2 = Map to security.provider
RestrictedSecurity(n).jce.provider.3 = Map to security.provider
RestrictedSecurity(n).jce.provider.4 = Map to security.provider

RestrictedSecurity(n).keystore.type = Map to keystore.type 
RestrictedSecurity(n).javax.net.ssl.keyStore = System property javax.net.ssl.keyStore

RestrictedSecurity(n).securerandom.provider = Secure random provider
RestrictedSecurity(n).securerandom.algorithm = Secure random algorithm
```

The new runtime command line arguments:

```
-Dsemeru.restrictedsecurity = (n)
    This flag will select the settings for the user specified fips policy.
-Dsemeru.restrictedsecurity= audit 
    This will list the name and number of all configured fips policies. it will NOT cause the jvm to terminate after printing the fips policies.
-Dsemeru.restrictedsecurity = trace
    This will list all properties relevant to the fips mode, including the existing default properties and the FIPS restrictions.
-Dsemeru.restrictedsecurity = help
    This will print help message.
```

The options can be combined: e.g. -Dsemeru.restrictedsecurity=1,trace,audit

During the process of reading the java.security properties file, the JDK will detect if Restricted Security mode as selected and the selected policy. This detection will be done as a check in the JCA code to see if the flag "semeru.restrictedsecurity" was used, if it is used, the specific policy number will be extracted, the code will then check the properties file to see if such a policy number exist. If a valid policy is selected, two configuration steps will occur:

1) JSSE Configuration during start up. If a valid policy was selected, the properties in java.security file will be read and mapped during the start up.

2) Cryptography available. The existing providers will be removed and only the compliant providers (defined by “SecurityRestrict(n).jce.provider”) will be added. And also, a list of permitted services that each provider may provide. A provider without any additional constraints will have all its services enabled. A provider with the constraints will only register the allowed constraints. For example:

```
RestrictedSecurity1.jce.provider.3 = SunEC [{KeyFactory, EC, ImplementedIn=Software: \
SupportedKeyClasses=java.security.interfaces.ECPublicKey|java.security.interfaces.ECPrivateKey: \
KeySize=256}, {AlgorithmParameters, EC, *}]
```


alon-sh:
thanks @taoliult - please provide a summary paragraph as to the content of this commit

alon-sh:
fyi @mstoodle

taoliult:
@alon-sh The summary paragraph added.

alon-sh:
@keithc-ca please review, updating FIPS code for both short and long term goals

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -1372,6 +1381,12 @@ protected void putService(Service s) {
             throw new IllegalArgumentException
                     ("service.getProvider() must match this Provider object");
         }
+        if (SecurityRestrictConfigurator.enableSecurityRestrict()) {
+            // If security restrict is not allow, then return without register
+            if (!SecurityRestrictProperties.getInstance().isConstraintsAllow(s)) {
```
keithc-ca, line 1386: I think we can find better names for the classes and methods so this reads more clearly. I suggest:
* SecurityRestrictConfigurator -> RestrictedSecurityConfigurator
* SecurityRestrictProperties -> RestrictedSecurityProperties
* enableSecurityRestrict -> isEnabled
* isConstraintsAllow -> isServiceAllowed

The comment should have a period at the end of the sentence, 

> // If security restrictions do not allow the service, then return without registering.

A better name for `s` would be `service`.

taoliult: @keithc-ca 
Thanks for the review. I updated the codes according to your review and also make other changes to make methods read more clearly. Please help to review.

@alon-sh  fyi

taoliult:
@keithc-ca @alon-sh 
I updated the codes according to the review and also make other changes to make method names read more clearly. Please help to review.

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
```diff
@@ -0,0 +1,320 @@
...
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, see <http://www.gnu.org/licenses/>.
+ *
+ * ===========================================================================
+ */
+
```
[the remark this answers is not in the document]
taoliult: @keithc-ca Which blanks line should I remove? Line 21 or line 24?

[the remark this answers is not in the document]
keithc-ca: I was referring to instances like line 24 (in this and other `.java` files), but if you want to also delete line 21 that is fine by me.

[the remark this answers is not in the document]
taoliult: @keithc-ca Thanks for the reply. I checked some other java files, I saw there is a blank line between the last line of copyright "*/" and the package declarations, same as above line 24. So, do we still need to remove this blank line 24?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,658 @@
...
+            throw new RuntimeException(
+                    "Restricted security mode initialization error, call getInstance with variables first.");
+        }
+
+        try {
+
+            // Print out the Help and Audit info.
+            if (userSecurityHelp) {
+                printHelp();
+                if ("0".equals(userSecurityNum)) {
```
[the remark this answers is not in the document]
taoliult: @keithc-ca 
I use the “0” as the default value of userSecurityNum. If the user did not specify the userSecurityNum, for example “-Dsemeru.restrictedsecurity=help”, then the default userSecurityNum value will be “0”. And that means, the user just want to print the "help" info, or the "audit" info, they did not want to run any codes in FIPS/RestrictedSecurity mode. 
For this, any better idea?

@alon-sh fyi.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,658 @@
...
+                providerInfo = providerInfo.trim().replaceAll(" ", "");
+                String[] inputArray = providerInfo.substring(providerInfo.indexOf("[") + 2, providerInfo.length() - 2)
+                        .split("\\},\\{");
+
+                // Column is type, algorithm and attributes.
+                String[][] constraints = new String[inputArray.length][3];
+
+                int cNum = 0;
+                for (String input : inputArray) {
+                    String[] constraint = input.trim().split(",");
```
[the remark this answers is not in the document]
taoliult: I change the codes as below(not check in yet):

```
                int cNum = 0;
                for (String input : inputArray) {
                    String[] InConstraint = input.trim().split(",");

                    int dNum = 0;
                    for (String InConsDetail : InConstraint) {
                        constraints[cNum][dNum] = notNullEmpty(InConsDetail) ? InConstraint.trim() : "*";
                        dNum ++;
                    }
                    for (int i = dNum; i < 3; i ++) {
                        constraints[cNum][dNum] = "*";
                    }
                    cNum++;
                }
```

If the input doesn't have exactly three parts, the "*" will be set to those missing parts.

src/java.base/share/classes/java/util/ServiceLoader.java
```diff
@@ -777,7 +786,18 @@ private S newInstance() {
             Throwable exc = null;
             if (acc == null) {
                 try {
-                    p = ctor.newInstance();
+                    // If the specified class extends Provider && restricted security mode enable.
+                    if (java.security.Provider.class.isAssignableFrom(ctor.getDeclaringClass())
+                            && RestrictedSecurityConfigurator.isEnabled()) {
```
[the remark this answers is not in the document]
taoliult: @keithc-ca Yes, this check should also be added in "if acc != null" and in "invokeFactoryMethod()". Added and testing right now before check in the code.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,658 @@
...
+            if (debug != null) {
+                debug.println("Loading constraints for security provider: " + providerName);
+            }
+
+            // Provider with constraints.
+            if (providerInfo.indexOf("[") > 0) {
+
+                providerInfo = providerInfo.trim().replaceAll(" ", "");
+                String[] inputArray = providerInfo.substring(providerInfo.indexOf("[") + 2, providerInfo.length() - 2)
+                        .split("\\},\\{");
```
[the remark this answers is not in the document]
taoliult: @keithc-ca

I changed the code as below by getting the index of “[{“. And this “[{“ must exist in the providerInfo if the providerInfo format correct. And after “[{“, there are characters or brackets if the providerInfo format correct, the format is verified in the method areBracketsBalanced(). So, +2 won’t cause any exception.

```
String[] inputArray = providerInfo.substring(providerInfo.indexOf("[{") + 2, providerInfo.length() - 2)
                        .split("\\},\\{");
```

@alon-sh fyi.

taoliult:
@keithc-ca  The codes updated according to the review suggestions. And I also replied those questions. Please help to review and advise.
@alon-sh  fyi.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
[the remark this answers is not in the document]
keithc-ca: Yes, please remove line 24 (you'll notice there isn't a corresponding blank line in `RestrictedSecurityProperties.java` which is also added in this pull request).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
```diff
@@ -0,0 +1,321 @@
...
+    private static final Debug debug = Debug.getInstance("semerufips");
+
+    // Restricted security mode enable check, only supported on Linux x64.
+    private static final boolean userEnabledSecurity;
+    private static final boolean isSecuritySupported;
+    private static final boolean shouldEnableSecurity;
+    private static final String userSecuritySetting;
+    private static final boolean userEnabledFIPS;
+
+    private static String userSecurityNum = "0";
```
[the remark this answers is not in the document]
keithc-ca: I don't think that's sufficient motivation to use `String` instead of `int`. The difference would just be use of `StringBuilder.append(int)` instead of `StringBuilder.append(String)`; if we were concerned about performance here, we might extract the common subexpression `"RestrictedSecurity" + userSecurityNum + ".desc`.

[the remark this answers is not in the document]
taoliult: @keithc-ca 
Thanks for the advice, made the changes and will check in the code later with other changes.

src/java.base/share/classes/java/util/ServiceLoader.java
```diff
@@ -884,6 +893,16 @@ private Provider<S> loadProvider(ServiceProvider provider) {
 
                 @SuppressWarnings("unchecked")
                 Class<? extends S> type = (Class<? extends S>) returnType;
+                // If the restricted security mode is enabled
+                // && the specified class extends java.security.Provider
+                // && the provider is NOT allowed in restricted security mode.
+                if (RestrictedSecurityConfigurator.isEnabled()
+                        && java.security.Provider.class.isAssignableFrom(clazz)
+                        && !RestrictedSecurityProperties.getInstance()
+                                .isProviderAllowed(clazz.getName())) {
```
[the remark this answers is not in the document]
taoliult: @keithc-ca 

In the ProviderConfig.java, it will also call this isProviderAllowed() method, to check if the java.base provider is allowed or not. Because the java.base provider will be initialized directly, not by the service loader. And at that place, it use the provider name, not the class.

So, I added an overload method with Class object as the parameter in the RestrictedSecurityProperties.java, to first check if the class extends java.security.Provider or not. If not, return false. If yes, call the isProviderAllowed() method with String as parameter to continue check if the provider is allowed or not.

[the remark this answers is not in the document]
taoliult: @keithc-ca 
Grinder openjdk tests failed after moving the isAssignable() into the isProviderAllowed() method. Working on it and will update shortly, mark this unresolved again.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
```diff
@@ -0,0 +1,322 @@
...
+                                System.getProperty("os.arch") };
+                    }
+                });
+        userEnabledFIPS = Boolean.parseBoolean(props[0]);
+        // If semeru.fips is true, then ignore semeru.restrictedsecurity, use userSecurityNum 1.
+        userSecuritySetting = userEnabledFIPS ? "1" : props[1];
+        userEnabledSecurity = !isNullOrBlank(userSecuritySetting);
+        isSecuritySupported = "Linux".equalsIgnoreCase(props[2])
+                && supportPlatforms.contains(props[3]);
+        shouldEnableSecurity = (userEnabledFIPS || userEnabledSecurity) && isSecuritySupported;
```
ZainabF92, line 80: I don't think `userEnabledFIPS` needs to be a field, since if `userEnabledFIPS` is `true` then `userEnabledSecurity` will definitely be `true`, because `userSecuritySetting` is set to `1`. This line can be
```
shouldEnableSecurity = userEnabledSecurity && isSecuritySupported;
```
and line 108 can be
```
if (userEnabledSecurity && !isSecuritySupported) {
```

taoliult: @ZainabF92
 
Hi Zainab,

The flag "userEnabledFIPS" is for back compatibility to the last version of FIPS mode code, which we released at this July release and also published the user guide. So, we have to keep this flag there even it doesn't needed for the restricted security mode.

taoliult:
@keithc-ca The codes updated according to the review suggestions. Please help to review and advise.

taoliult:
@keithc-ca Any update for the code review? Thank.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,697 @@
...
+            }
+
+            int pos = providerInfo.indexOf('[');
+            String providerName = (pos < 0) ? providerInfo.trim() : providerInfo.substring(0, pos).trim();
+            // Provider with argument (provider name + optional argument).
+            providers.add(pNum - 1, providerName);
+
+            pos = providerName.indexOf(' ');
+            providerName = (pos < 0) ? providerName.trim() : providerName.substring(0, pos).trim();
+            // Provider without argument.
```
ZainabF92, line 220: These lines need to be added here as well: the package should not be part of the provider’s “short name”, otherwise the comparison on line 433 fails.
```
// Remove the provider's class package name.
pos = providerName.lastIndexOf('.');
providerName = (pos > 0) ? providerName.substring(pos + 1, providerName.length()) : providerName;
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
```diff
@@ -0,0 +1,325 @@
...
+                    new RuntimeException("user restricted security setting " + userSecuritySetting + " incorrect.")
+                            .printStackTrace();
+                    System.exit(1);
+                }
+            }
+        }
+
+        if (debug != null) {
+            debug.println("Loaded user restricted security settings, with userSecurityNum: "
+                    + String.valueOf(userSecurityNum) + " userSecurityTrace: " + userSecurityTrace
```
keithc-ca, line 219: The use of `String.valueOf()` is unnecessary here. Putting one parameter per line makes this easier to read:
```java
            debug.println("Loaded user restricted security settings, with userSecurityNum: " + userSecurityNum
                    + " userSecurityTrace: " + userSecurityTrace
                    + " userSecurityAudit: " + userSecurityAudit
                    + " userSecurityHelp: " + userSecurityHelp);
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
```diff
@@ -0,0 +1,325 @@
...
+                    + " userSecurityAudit: " + userSecurityAudit + " userSecurityHelp: " + userSecurityHelp);
+        }
+    }
+
+    /**
+     * Add restricted security providers.
+     *
+     * @param providers the provider name array
+     */
+    private static void setProviders(Properties props, ArrayList<String> providers) {
```
keithc-ca, line 229: Please use the more generic `List` instead of `ArrayList`, even though this is a private method.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
```diff
@@ -0,0 +1,325 @@
...
+
+        if (debug != null) {
+            debug.println("Adding restricted security provider.");
+        }
+
+        int pNum = 1;
+        for (String provider : providers) {
+            props.setProperty("security.provider." + pNum, provider);
+            pNum ++;
```
keithc-ca, line 238: Rather than start at 1, just increment before the use:
```java
        int pNum = 0;
        for (String provider : providers) {
            pNum += 1;
            props.setProperty("security.provider." + pNum, provider);
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
```diff
@@ -0,0 +1,325 @@
...
+     */
+    private static void setProperties(Properties props, RestrictedSecurityProperties properties) {
+
+        if (debug != null) {
+            debug.println("Adding restricted security properties.");
+        }
+
+        Map<String, String> propsMapping = new HashMap<>();
+
+        // JDK properties name as Key, restricted security properties vaule as value.
```
keithc-ca, line 258: vaule -> value

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
```diff
@@ -0,0 +1,325 @@
...
+        propsMapping.put("jdk.tls.ephemeralDHKeySize", properties.getJdkTlsDphemeralDHKeySize());
+        propsMapping.put("jdk.tls.legacyAlgorithms", properties.getJdkTlsLegacyAlgorithms());
+        propsMapping.put("jdk.certpath.disabledAlgorithms", properties.getJdkCertpathDisabledAlgorithms());
+        propsMapping.put("jdk.security.legacyAlgorithm", properties.getJdkSecurityLegacyAlgorithm());
+
+        for (Map.Entry<String, String> entry : propsMapping.entrySet()) {
+            String jdkPropsName = entry.getKey();
+            String propsNewValue = entry.getValue();
+
+            String propsOldValue = !isNullOrBlank(props.getProperty(jdkPropsName)) ? props.getProperty(jdkPropsName) : "";
```
keithc-ca, line 270: Please extract the common subexpression `props.getProperty(jdkPropsName)`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
```diff
@@ -0,0 +1,325 @@
...
+                }
+            }
+        }
+
+        // For keyStore and keystore.type, old value not needed, just set the new value.
+        if (!isNullOrBlank(properties.getKeyStoreType())) {
+            props.setProperty("keystore.type", properties.getKeyStoreType());
+        }
+        if (!isNullOrBlank(properties.getKeyStore())) {
+            System.setProperty("javax.net.ssl.keyStore", properties.getKeyStore());
```
keithc-ca, line 287: Common subexpressions `properties.getKeyStoreType()`, `properties.getKeyStore()`.
Why does this use `System` instead of `props` on line 287?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
```diff
@@ -0,0 +1,325 @@
...
+     * @param descSunsetDate the sun set date from java.security
+     * @return true if the restricted security policy sunset
+     */
+    private static boolean isPolicySunset(String descSunsetDate) {
+
+        boolean isSunset = false;
+        try {
+            isSunset = LocalDate.parse(descSunsetDate, DateTimeFormatter.ofPattern("yyyy-MM-dd"))
+                    .isBefore(LocalDate.now());
+        } catch (Exception except) {
```
keithc-ca, line 303: This should catch the more specific exception `DateTimeParseException`.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
```diff
@@ -0,0 +1,325 @@
...
+            debug.println("Restricted security policy is sunset: " + isSunset);
+        }
+        return isSunset;
+    }
+
+    /**
+     * Check if the input string is null and empty.
+     *
+     * @param string the input string
+     * @return true if the input string is null and emtpy
```
keithc-ca, line 320: The text should say "string is null **or blank**" not "string is null **and empty**".

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,707 @@
...
+    // For Secure Random.
+    private static String jdkSecureRandomProvider;
+    private static String jdkSecureRandomAlgorithm;
+
+    // Provider with argument (provider name + optional argument).
+    private static ArrayList<String> providers = new ArrayList<String>();;
+    // Provider without argument.
+    private static ArrayList<String> providersSN = new ArrayList<String>();;
+    // Constraints for each provider. Key is the Provider Name, Value is the Constraints.
+    private static Map<String, String[][]> providerConstraints = new HashMap<String, String[][]>();
```
keithc-ca, line 68: Please define fields using interfaces (`List`, not `ArrayList`), avoid redundant type parameters and superfluous semicolons:
```java
    private static List<String> providers;
    // Provider without argument.
    private static List<String> providersSN;
    // Constraints for each provider if there are.
    private static Map<String, Constraint[]> providerConstraints;
```
Initialization should occur in the constructor.
Note also the use of a nested class:
```java
    private static final class Constraint {
        final String type;
        final String algorithm;
        final String attributes;

        Constraint(String type, String algorithm, String attributes) {
            super();
            this.type = type;
            this.algorithm = algorithm;
            this.attributes = attributes;
        }

    }
```

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,707 @@
...
+            String providerName = providersSN.get(pNum - 1);
+            String providerInfo = securityProps
+                    .getProperty(propsPrefix + ".jce.provider." + pNum);
+
+            if (debug != null) {
+                debug.println("Loading constraints for security provider: " + providerName);
+            }
+
+            // Remove all the space.
+            providerInfo = providerInfo.trim().replaceAll(" ", "");
```
keithc-ca, line 296: I think we want more sophisticated parsing; there may be other kinds of whitespace, for example, and we should be providing better feedback when the string is not well-formed.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,707 @@
...
+                // Column is type, algorithm and attributes.
+                String[][] constraints = new String[inputArray.length][3];
+
+                int cNum = 0;
+                for (String input : inputArray) {
+                    String[] inConstraint = input.trim().split(",");
+
+                    int dNum = 0;
+                    for (String inConsDetail : inConstraint) {
+                        constraints[cNum][dNum] = !isNullOrBlank(inConsDetail) ? inConsDetail.trim() : "*";
```
keithc-ca, line 312: This will fail if there are more than 3 details.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,707 @@
...
+                        }
+                        break;
+                    case ']':
+                        check = stack.pop();
+                        if (check == '(' || check == '{') {
+                            return false;
+                        }
+                        break;
+                }
+            } catch (NoSuchElementException ex) {
```
keithc-ca, line 640: An empty stack should be handled explicitly instead of via an exception.

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
taoliult: @keithc-ca 
Add more checks in the methods initConstraints() to provide better feedback when the string is not well-formed.

src/java.base/share/classes/sun/security/provider/SunEntries.java
```diff
@@ -114,266 +112,218 @@ public static List<String> createAliasesWithOid(String ... oids) {
...
+            add(p, "SecureRandom", "NativePRNGBlocking",
+                    "sun.security.provider.NativePRNG$Blocking", null, attrs);
+        }
+        if (NativePRNG.NonBlocking.isAvailable()) {
+            add(p, "SecureRandom", "NativePRNGNonBlocking",
+                    "sun.security.provider.NativePRNG$NonBlocking", null, attrs);
+        }
+        attrs.put("ImplementedIn", "Software");
+        add(p, "SecureRandom", "DRBG", "sun.security.provider.DRBG",
                 null, attrs);
```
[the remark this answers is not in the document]
taoliult: @keithc-ca 

For these example lines, I compared and did not see they are different than in https://github.com/ibmruntimes/openj9-openjdk-jdk11/commit/d9adaafe153c9d6e5826d1f7fa72e1e38c20e20c. For the SunEntries.java file, in the current commit, I removed all the FIPS checks made by commit https://github.com/ibmruntimes/openj9-openjdk-jdk11/commit/eaced24d8ce3b97bad578e11689f826c7ef9412b, because we don't need to check FIPS in SunEntries.java anymore. 

Could you help to provider some detailed examples about what changes need to revert to make them same as commit https://github.com/ibmruntimes/openj9-openjdk-jdk11/commit/d9adaafe153c9d6e5826d1f7fa72e1e38c20e20c ?

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
taoliult: @keithc-ca 

"keystore.type" is a property defined in the java.security file, so we can use props.setProperty() to set its value.

"javax.net.ssl.keyStore" is a SSL properties set at the JVM level via system properties. So, we have to either set them when running the program (java -D....), or set them in code by doing System.setProperty().

taoliult:
@keithc-ca 
The codes updated according to the review suggestions. Please help to review and advise. And I also added more checks in the methods initConstraints() to provide better feedback when the string is not well-formed.

taoliult:
@keithc-ca Any updates about the code review?

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
```diff
@@ -193,182 +191,128 @@ public Void run() {
...
-                names.append(namedCurve.getObjectId());
-                names.append("]");
+        ATTRS.put("ImplementedIn", "Software");
+        String ecKeyClasses = "java.security.interfaces.ECPublicKey" +
+                 "|java.security.interfaces.ECPrivateKey";
+        ATTRS.put("SupportedKeyClasses", ecKeyClasses);
+        ATTRS.put("KeySize", "256");
+
+        /*
+         * Key Factory engine
```
keithc-ca, line 201: Please revert unnecessary changes relative to 9494bb5ffb761591c7e04a8a4b5b3380bee05a91; here and on line 301.

keithc-ca (changes requested):

src/java.base/share/classes/sun/security/jca/ProviderConfig.java
```diff
@@ -162,6 +171,12 @@ public String toString() {
     // com.sun.net.ssl.internal.ssl.Provider has been deprecated since JDK 9
     @SuppressWarnings("deprecation")
     synchronized Provider getProvider() {
+        // If provider is not allowed in restricted security mode, return without load.
+        if (RestrictedSecurityConfigurator.isEnabled()) {
+            if (!RestrictedSecurityProperties.getInstance().isProviderAllowed(provName)) {
+                return null;
+            }
+        }
```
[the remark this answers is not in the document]
taoliult: @keithc-ca 
In this ProviderConfig.java file, after line 355, there was a "p == null" check I added, but I already removed it. Because the not-allowed provider checks are added in the ServiceLoader.java before calling the iterator hasNext() method. So the provider p returned by the iter.next() won't be null.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityConfigurator.java
taoliult: @keithc-ca 

"keystore.type" is a Java security property defined in the Java security properties file(the java.security file). The Security.java loads this properties file as a Properties object. so we can use Properties.setProperty() to set or change its value.

"javax.net.ssl.keyStore" is a SSL property, means the location of the Java keyStore file containing an application its own certificates and private keys. In the FIPS mode, it using the NSS database(when keystore.type set to PKCS11 format) to store the certificates and keys, so the keyStore file location must be set to NONE. That mean, we must set "javax.net.ssl.keyStore" to NONE. And due to the SSL properties set at the JVM level via system properties, not read by the codes as a Properties object. So, we have to either set them when running the program (java -D....), or set them in code by doing System.setProperty().

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
taoliult: @keithc-ca 

I changed the removal of all whitespaces, to just remove the whitespaces in the format separator if there are, like "[{", "}]" and "},{". So I can then use this separator to get the constraint by splitting the string.

taoliult:
@keithc-ca
The codes updated according to the review suggestions, and I also replied those questions. Please help to review and advise.

taoliult:
@keithc-ca
The codes updated according to the review suggestions, and I also replied those questions. Please help to review and advise.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,765 @@
...
+
+        Constraint[] constraints = providerConstraints.get(providerName);
+
+        // Go into the security provider constraints check if there are.
+        if (constraints != null && constraints.length > 0) {
+
+            for (Constraint constraint : constraints) {
+
+                boolean cTypePut = "*".equals(constraint.getType())
+                        || type.equals(constraint.getType());
```
keithc-ca, line 391: Repeated subexpressions (`constraint.getType()` in this case); suggest a helper method that checks for `"*"` or a specific string.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,765 @@
...
+                // For type and algorithm match, and attribute is not *.
+                // Then continue checking attributes.
+                if (cTypePut && cAlgorithmPut) {
+                    String[] cAttributes = constraint.getAttributes().split(":");
+
+                    // For each attribute, must be all matched for return allowed.
+                    for (String cAttribute : cAttributes) {
+                        String[] input = cAttribute.split("=");
+
+                        if (input.length != 2) {
```
keithc-ca, line 416: This test should be done at initialization time (not later when checking for allowed services).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,765 @@
...
+            return type;
+        }
+
+        String getAlgorithm() {
+            return algorithm;
+        }
+
+        String getAttributes() {
+            return attributes;
+        }
```
keithc-ca, line 699: Suggest removing these methods and simply access the fields directly (it's a private class).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,765 @@
...
+        return descName;
+    }
+
+    public String getDescNumber() {
+        return descNumber;
+    }
+
+    public String getDescPolicy() {
+        return descPolicy;
+    }
```
keithc-ca, line 712: Public functions need javadoc and these should have better names (at least expand "Desc" to "Description" or similar).

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,765 @@
...
+        return jdkSecureRandomProvider;
+    }
+
+    public String getJdkSecureRandomAlgorithm() {
+        return jdkSecureRandomAlgorithm;
+    }
+
+    public Map<String, Constraint[]> getProviderConstraints() {
+        return providerConstraints;
+    }
```
keithc-ca, line 764: This can't be much use outside this class: it involves a private class.

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
taoliult: Remove this method since no outside this class invoking needed.

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurityProperties.java
```diff
@@ -0,0 +1,765 @@
...
+
+        try {
+            // Print out the Help and Audit info.
+            if (userSecurityHelp) {
+                printHelp();
+                if (userSecurityNum == 0) {
+                    if (debug != null) {
+                        debug.println("Print out the help info and exit.");
+                    }
+                    System.exit(0);
```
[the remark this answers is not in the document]
taoliult: @keithc-ca 
When userSecurityNum == 0, that mean from the user input, there is no security number specified. That also mean, the user only want to print out the help or audit info, but not running the application. So I added an exit() here if the userSecurityNum == 0.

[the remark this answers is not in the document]
taoliult: And I changed these codes to add the check for "userSecurityTrace" when no security number specified, to print out more useful info when userSecurityNum is 0 and userSecurityTrace is true.

taoliult:
@keithc-ca
The codes updated according to the review suggestions. Please help to review and advise.

JasonFengJ9:
The following code snippet shows that a security provider could be instantiated before this restricted security mode is in effect
```
Provider p = new sun.security.ec.SunEC();
System.out.println("p = " + p);
Service s = p.getService("KeyAgreement", "ECDH");
System.out.println("s = " + s);
...
```
the output
```
p = SunEC version 11
s = SunEC: KeyAgreement.ECDH -> sun.security.ec.ECDHKeyAgreement
  attributes: {ImplementedIn=Software, KeySize=256, SupportedKeyClasses=java.security.interfaces.ECPublicKey|java.security.interfaces.ECPrivateKey}
...
semerufips: Restricted security mode detected, loading...
```

@taoliult does this indicate a security provider usage escaped the restricted security mode?

keithc-ca:
That sample requires bypassing modularity boundaries:
```
		Provider p = new sun.security.ec.SunEC();
		                             ^
  (package sun.security.ec is declared in module jdk.crypto.ec, which does not export it)
```
but it raises a good point.

@taoliult What is the plan to address this?

taoliult:
@keithc-ca @JasonFengJ9 
Working with Jason in slack for this issue, will update once find anything.

taoliult:
@keithc-ca @JasonFengJ9 

In the restricted security mode design, the restricted security leverage the JCA Security class, which manages the installed providers and security-wide properties. All the restricted security settings initialized starting from this Security class. The providers, which we defined in the java.security file with its constraints, will be initialized and got by invoking the Security.getProvider() method.

When invoke the Security class Security.getProvider() method, the static initialize() method in Security.java will be invoked, and the java.security system properties file will be loaded. And then, go into the RestrictedSecurityConfigurator.configure() method to load and initialize the restricted security mode.

```
        System.out.println("Start new sun.security.ec.SunEC()");
        Provider pSunEC = new sun.security.ec.SunEC();
        dumpSunECService(pSunEC);
```

```
        void dumpSunECService(Provider p) {
                System.out.println("dumpSunECService() p = " + p);
                if (p != null) {
                        Service s = p.getService("KeyAgreement", "ECDH");
                        System.out.println("s = " + s);
                        System.out.println("s.getAlgorithm(): " + s.getAlgorithm());
                        System.out.println("s.getClassName(): " + s.getClassName());
                        System.out.println("s.getClass(): " + s.getClass());
                }
        }
```

```
        System.out.println("_______________");
        {
                System.out.println("Start Security.getProvider");
                Provider p = Security.getProvider("SunEC");
                dumpSunECService(p);
        }
```

In the above example codes, it first initialize a SunEC provider by calling its construction method directly, get its services and print them out. These happens before calling the Security.getProvider(). That mean, these happens before initializing the Security class and FIPS.

My understanding is, if the application wants to use the java security, the installed FIPS security provider and its security properties and constraints, the provider should be initialized by invoking the Security.getProvider() method, not by invoking its construction method directly.

JasonFengJ9:
> My understanding is, if the application wants to use the java security, the installed FIPS security provider and its security properties and constraints, the provider should be initialized by invoking the Security.getProvider() method, not by invoking its construction method directly.

How about `semeru.restrictedsecurity`? does this have `Security.getProvider()` limitation as well?

taoliult:
@JasonFengJ9 

Yes, the "-Dsemeru.fips=true" is only used for FIPS, we apply this property in our last release of FIPS. In this current codes, we extend it to Java Restricted Security Mode, adding a new property "semeru.restrictedsecurity". So, according to the policy defined in the java.security file, it can apply not only for FIPS, but also apply to other restriction if needed, for example the CRIU. 

Right now, the policy number 1 is "RestrictedSecurity1.desc.name = Red Hat Enterprise Linux 8 NSS Cryptographic Module FIPS 140-2". This is for FIPS mode. There may be other policy in the future.

So, "-Dsemeru.restrictedsecurity=1" and "-Dsemeru.fips=true", there are same. We still keep the "-Dsemeru.fips=true" property is because, the last release of FIPS and its user guide mentioned this property to enable FIPS, so we have to keep this for the back compatibility.

taoliult:
JDKnext PR https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/544 created.

JasonFengJ9:
> So, according to the policy defined in the java.security file, it can apply not only for FIPS, but also apply to other restriction if needed, for example the CRIU.

I was more concerned about `semeru.restrictedsecurity` for CRIU usage, it seems CRIU specific support is required since there is no restriction for direct provider/service usage.

On the other hand, does the FIPS mode handle a case like
https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/9496bc7ede8e4caec06c68e444beeb5199fbd1d4/src/java.base/share/classes/java/io/File.java#L1955

taoliult:
@keithc-ca 

This PR on JDK11 is ready. It has all the code review updates from JDKNext PR https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/544

@jasonkatonica @WilburZjh FYI.

keithc-ca:
I will consider this after equivalent changes have been merged into jdk20, jdk19 and jdk17 (in that order).

pshipton:
I think we should skip jdk19, I don't plan to add the change to the 0.37 release branch, and the jdk19 head stream won't be used for another release.

keithc-ca:
> we should skip jdk19

That works for me.

src/java.base/share/classes/sun/security/jca/ProviderConfig.java
```diff
@@ -39,6 +39,8 @@
 
 import sun.security.util.PropertyExpander;
 
+import openj9.internal.security.RestrictedSecurity;
```
keithc-ca, line 42: The first year in the copyright notice for this file should be 2022 as that is when this work began.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
keithc-ca: Those changes have not been reverted.

keithc-ca (changes requested):

taoliult:
@keithc-ca 
Codes updated according to review suggestion. Please help to review and advise. Thanks.

src/jdk.crypto.ec/share/classes/sun/security/ec/SunEC.java
```diff
@@ -193,182 +191,128 @@ public Void run() {
...
+             "sun.security.ec.ECDSASignature$SHA256inP1363Format"));
+        putService(new ProviderService(this, "Signature",
+            "SHA384withECDSAinP1363Format",
+            "sun.security.ec.ECDSASignature$SHA384inP1363Format"));
+        putService(new ProviderService(this, "Signature",
+            "SHA512withECDSAinP1363Format",
+            "sun.security.ec.ECDSASignature$SHA512inP1363Format"));
+
+        /*
+         * Key Pair Generator engine
```
keithc-ca, line 301: This line should match line 283 in 9494bb5ffb761591c7e04a8a4b5b3380bee05a91; there's no need for it to be different in this branch.

keithc-ca:

PR #596, opened 2022-12-05
Enable CRIU on x86

tajila:
Enable CRIU on x86

Signed-off-by: Tobi Ajila <atobia@ca.ibm.com>

### DIFF
--- a/closed/autoconf/custom-hook.m4
+++ b/closed/autoconf/custom-hook.m4
@@ -376,7 +376,15 @@ AC_DEFUN([OPENJ9_CONFIGURE_CRIU_SUPPORT],
   elif test "x$enable_criu_support" = xno ; then
     AC_MSG_RESULT([no (explicitly disabled)])
   elif test "x$enable_criu_support" = x ; then
-    AC_MSG_RESULT([no (default)])
+    case "$OPENJ9_PLATFORM_CODE" in
+      xa64)
+        AC_MSG_RESULT([yes (default)])
+        OPENJ9_ENABLE_CRIU_SUPPORT=true
+        ;;
+      *)
+        AC_MSG_RESULT([no (default)])
+        ;;
+    esac
   else
     AC_MSG_ERROR([--enable-criu-support accepts no argument])
   fi


tajila:
jenkins test sanity xlinux jdk11

tajila:
jenkins test sanity xlinux jdk11

tajila:
@keithc-ca please review these changes

tajila:
jenkins test sanity xlinux jdk11

pshipton:
CRIU testing is failing on [cent6-x64-6](https://openj9-jenkins.osuosl.org/computer/cent6-x64-6)
`org.eclipse.openj9.criu.SystemCheckpointException: The JVM attempted to load libcriu.so but was unable to: 1`

@tajila @AdamBrousseau is this going to be a problem on all the cent6 machines?

AdamBrousseau:
Tobi can confirm, but I believe so yes. Once CRIU is needed by default for a test target, that target will no longer be able to be run on cent/rhel6.

pshipton:
Seems we don't need the cent6 machines for building. I'm leaning towards just disabling these machines for now, and when we have time we can upgrade them to a different OS.

pshipton:
We have 20 xlinux machines, 7 of which are cent.6. Not sure what disabling them will do to the builds. I'll do it now and we'll see if there are any problems in the nightly builds tonight.

pshipton:
The other option that avoids having to get rid of the cent6 machines is to fix the CRIU tests to detect cent6 and don't try to run on this platform.

pshipton:
Even if we decide to keep them, we don't need 7. I'd keep 3 and upgrade the rest.

llxia:
> Is there a way to exclude the criu tests on cent6 machines?

Quick way: CRIU-specific tests are only in sanity.functional and sanity.external. The rest test targets do not have CRIU-specific tests. Once CRIU is needed by default on x86, we can only add `!sw.os.cent.6` for sanity.functional and sanity.external for now.

As for excluding the CRIU tests on cent6 machines, it needs some work. We should be able to leverage TKG micro architecture detection. We need to add logic in TKG to detect cent6, then update `platformRequirements` for each CRIU test.

On a related note, currently, CRIU tests have to run on the machines that have CRIU installed. By enabling CRIU feature as default on x86, does it mean we will install CRIU on all x86 test machines? If not, the CRIU tests will fail on the machine without `ci.role.test.criu`.

AdamBrousseau:
>does it mean we will install CRIU on all x86 test machines?

That is the current plan. If we can install on all test machines, then a label won't be required. TBD.

pshipton:
No problems in the nightly builds after disabling the cent6 machines.

Re-run the failed targets https://openj9-jenkins.osuosl.org/job/Grinder/1710, which passed.

pshipton (approved):

PR #601, opened 2022-12-08
Merge jdk-11.0.18+5 and jdk-11.0.18+6 to 0.36

pshipton:
Depends on https://github.com/eclipse-openj9/openj9/pull/16441 which is already merged.


JasonFengJ9 (approved):
The latest abuild seems green, no need to wait for its promotion into `openj9` branch, using `openj9-staging` looks good.

PR #608, opened 2023-01-05
Fix problem building openssl when ccache is enabled

keithc-ca:
This is a back-port of https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/536.


pshipton:
jenkins compile all jdk11

pshipton (approved):

PR #614, opened 2023-01-19
Ensure test-image-openj9 target fails when java -version fails

jdmpapin:
Previously, the exit status could come from `tee`, which succeeds.

There is logic in `make/autoconf/basics.m4` that checks for support and then opportunistically adds `-o pipefail` to `SHELL` in the makefiles, which already accomplishes the same thing, but only when `pipefail` support is detected. That should be most of the time, but it's not obvious that we should expect `pipefail` support in all cases.

With this change, the target will fail whenever `java -version` fails, regardless of the `pipefail` setting.

### DIFF
--- a/closed/custom/Main.gmk
+++ b/closed/custom/Main.gmk
@@ -1,5 +1,5 @@
 # ===========================================================================
-# (c) Copyright IBM Corp. 2017, 2021 All Rights Reserved
+# (c) Copyright IBM Corp. 2017, 2023 All Rights Reserved
 # ===========================================================================
 # This code is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License version 2 only, as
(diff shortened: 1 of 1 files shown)


closed/custom/Main.gmk
```diff
@@ -76,7 +76,7 @@ ALL_TARGETS += debug-image
 test-image-openj9 : exploded-image
 	@+$(OPENJ9_MAKE) openj9_test_image
 ifneq ($(COMPILE_TYPE), cross)
-	$(JDK_OUTPUTDIR)/bin/java -version 2>&1 | $(TEE) $(TEST_IMAGE_DIR)/openj9/java-version.txt
+	rc=0; $(JDK_OUTPUTDIR)/bin/java -version >$(TEST_IMAGE_DIR)/openj9/java-version.txt 2>&1 || rc=$$?; $(CAT) $(TEST_IMAGE_DIR)/openj9/java-version.txt; exit "$$rc"
```
keithc-ca, line 79: Why the difference as compared with https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/639?
It's unclear to me how this is better, but in any case, we should be consistent.

jdmpapin: It's because `basics.m4` also opportunistically sets `-e` in `SHELL`. If that happens to be set, then as soon as `java -version` fails, the shell exits and we don't get to run `cat`, so the output is hidden.

jdmpapin: That's not necessary in Java 8 because its makefile doesn't try to pass `-e` to the shell. Though if you like, I can change the Java 8 PR to be consistent with these others anyway.

keithc-ca: At least on Linux, `spec.gmk` for jdk11+ says:
```
BASH_ARGS:= -o pipefail -e
```
so this isn't needed; and perhaps it's reasonable to assume `bash` in jdk8 so we could use `-o pipefail` there.

keithc-ca: I looked at the configuration code some more. It seems to anticipate the possibility that `pipefail` is not supported by the available version of `bash`, so we _do_ want this change (and the jdk8 change should match this).

jdmpapin:
Added quotes for consistency

keithc-ca:
Jenkins compile alinux64 jdk11

keithc-ca (approved):

PR #616, opened 2023-02-01
Correct IBM copyright date

keithc-ca:
Changes were made internally in 2021.


pshipton:
In what commit was that?

keithc-ca:
I was mistaken: The earliest internal commit I could find is 04414c146f9e99a3cc8ed1e7ad6f0310c9c48582 which was authored 2022-06-10.

PR #619, opened 2023-02-14
OpenJ9 doesn't support Perf, don't try to use it

pshipton:
Trying to use jdk.internal.perf.Perf may result in an Assertion "Java_sun_misc_Perf_attach unimplemented" or similar. Calling sun.tools.jconsole.LocalVirtualMachine.getAllVirtualMachines() tries to use Perf.

Issue https://github.com/eclipse-openj9/openj9/issues/16719

I didn't add any IBM copyright since no code was added.

Will port to other versions after the initial review.

### DIFF
--- a/src/jdk.internal.jvmstat/share/classes/sun/jvmstat/PlatformSupport.java
+++ b/src/jdk.internal.jvmstat/share/classes/sun/jvmstat/PlatformSupport.java
@@ -56,16 +56,7 @@ public class PlatformSupport {
     }
 
     public static PlatformSupport getInstance() {
-        try {
-            Class<?> c = Class.forName("sun.jvmstat.PlatformSupportImpl");
-            @SuppressWarnings("unchecked")
-            Constructor<PlatformSupport> cntr = (Constructor<PlatformSupport>) c.getConstructor();
-            return cntr.newInstance();
-        } catch (ClassNotFoundException e) {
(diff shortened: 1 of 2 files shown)


src/jdk.internal.jvmstat/share/classes/sun/jvmstat/PlatformSupport.java
```diff
@@ -56,16 +56,7 @@ public class PlatformSupport {
...
-            Class<?> c = Class.forName("sun.jvmstat.PlatformSupportImpl");
-            @SuppressWarnings("unchecked")
-            Constructor<PlatformSupport> cntr = (Constructor<PlatformSupport>) c.getConstructor();
-            return cntr.newInstance();
-        } catch (ClassNotFoundException e) {
-            return new PlatformSupport();
-        } catch (ReflectiveOperationException e) {
-            throw new InternalError(e);
-        }
+       return new PlatformSupport();
```
keithc-ca, line 59: Please correct the indentation here.

pshipton: Updated

keithc-ca (approved):
Same change as tested in ibmruntimes/openj9-openjdk-jdk20#17.

PR #620, opened 2023-02-17
Merge master jdk-11.0.19+2 into openj9-staging

JasonFengJ9:
Resolved conflicts:
```
	make/autoconf/build-aux/config.guess
	make/lib/Lib-java.base.gmk
```


make/autoconf/build-aux/config.guess
```diff
@@ -33,7 +33,40 @@
 # ===========================================================================
 
 DIR=`dirname $0`
-OUT=`. $DIR/autoconf-config.guess`
+OUT=`. $DIR/autoconf-config.guess 2> /dev/null`
+
+# Handle some cases that autoconf-config.guess is not capable of
+if [ "x$OUT" = x ]; then
```
keithc-ca, line 39: With these changes, I don't think we need any changes to this file as compared with upstream.

keithc-ca (changes requested):

make/autoconf/build-aux/config.guess
JasonFengJ9: Remove the OpenJ9 patch.

pshipton:
There is still a difference to the OpenJDK version
```
if test $? != 0; then
  # The GNU version of uname may be on the PATH which reports arm64 instead
  echo $OUT | grep arm64-apple-darwin > /dev/null 2> /dev/null
fi
```

make/autoconf/build-aux/config.guess
pshipton: There is still a difference to the OpenJDK version
```
if test $? != 0; then
  # The GNU version of uname may be on the PATH which reports arm64 instead
  echo $OUT | grep arm64-apple-darwin > /dev/null 2> /dev/null
fi
```

pshipton: Although jdk17 has that code.

pshipton: If we're going to backport this change, which is fine with me, I think it should be done in a separate commit and not part of a merge commit.

JasonFengJ9: In the past, if there was a merge conflict, and the conflict can be resolved by removing the OpenJ9 patch, the change was made within the merge commit for convenience.

pshipton: But you didn't remove the OpenJ9 patch entirely. There are two ways we could look at it, either you removed only part of the OpenJ9 patch, or you removed all of it and backported a change from OpenJDK 17. I suggest the second way is the better way to view it. You can remove the OpenJ9 change entirely in the merge commit, but backporting the change from 17 should have it's own commit.

JasonFengJ9: > But you didn't remove the OpenJ9 patch entirely. 

Just pushed the change.

pshipton: lgtm now. We don't have the code I copied above, and we risk breaking the build for us or somebody else, but if we find the code is necessary it can be backported in a new commit.

pshipton:
jenkins compile amac jdk11

make/autoconf/build-aux/config.guess
keithc-ca: We shouldn't need the code quoted above; [this line from upstream](https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/master/make/autoconf/build-aux/config.guess#L128) should map `arm-apple-darwin` or `arm64-apple-darwin` to `aarch64-apple-darwin` when appropriate.

keithc-ca (approved):

pshipton (approved):

PR #626, opened 2023-03-15
Fix memory leaks related to openssl EC pathways

jasonkatonica:
This update fixes a few memory leaks associated with the creation of 
EC private and public keys.

This openssl api BN_bn2bin described here:

https://www.openssl.org/docs/man1.0.2/man3/BN_bin2bn.html

This api states that a new BIGNUM is created if the last argument 
is null to the API. In our case the last argument is indeed null 
so we are leaking memory in the EC paths since we never free the big 
number returned.

A cleanup goto was also added to simplify the code and ensure that
all cleanup of allocated memory is done. We make explicit calls
to then free all allocated memory in the cleanup.

This is a cherry pick of:
https://github.com/ibmruntimes/openj9-openjdk-jdk/commit/969654a4805f84934890dd838ad8ecd346e47c12

Signed-off-by: Jason Katonica katonica@us.ibm.com

### DIFF
--- a/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
+++ b/closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
@@ -2340,6 +2340,14 @@ Java_jdk_crypto_jniprovider_NativeCrypto_ECCreatePublicKey
     (*env)->ReleasePrimitiveArrayCritical(env, x, nativeX, JNI_ABORT);
     (*env)->ReleasePrimitiveArrayCritical(env, y, nativeY, JNI_ABORT);
 
+    if (NULL != xBN) {
+        (*OSSL_BN_free)(xBN);
+    }
+
+    if (NULL != yBN) {
+        (*OSSL_BN_free)(yBN);
+    }
+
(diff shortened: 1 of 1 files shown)


jasonkatonica:
This fix is associated with the issue located at https://github.com/eclipse-openj9/openj9/issues/16451

keithc-ca:
Please update the commit message to follow these [guidelines](https://github.com/eclipse/omr/blob/master/CONTRIBUTING.md#commit-guidelines), in particular the suggested line-length limit.

keithc-ca:
Your commit should list you as the author, not me.

keithc-ca:
For the record, the build was successful (https://openj9-jenkins.osuosl.org/job/PullRequest-OpenJDK11/115/); I'm not anticipating any upcoming code changes so that result will stand.

jasonkatonica:
Updated the PR text and commit text to follow the 72 character boundary where possible.

Since I cherry picked this from a merged commit, the author was listed as both Keith and Myself. I changed the author to only myself.

keithc-ca (approved):

PR #635, opened 2023-03-23
Backport missing Oracle CPE from jdk17

pshipton:
The files are missing the Oracle "Classpath" exception in jdk11 but are otherwise identical to jdk17.

We will request OpenJDK to fix this for future releases.


keithc-ca (approved):

keithc-ca:
No build needed to verify comment changes.

PR #640, opened 2023-04-05
Allow any text without whitespace for the openssl version

pshipton:
Will port to other versions after review.

Example build using this https://openj9-jenkins.osuosl.org/job/Build_JDK11_aarch64_mac_Personal/316

  extra_getsource_options: '--openssl-version=OpenSSL_1_1_1-stable --openssl-repo=https://github.com/ibmruntimes/openssl.git'

### DIFF
--- a/closed/get_openssl_source.sh
+++ b/closed/get_openssl_source.sh
@@ -1,7 +1,7 @@
 #!/bin/sh
 
 # ===========================================================================
-# (c) Copyright IBM Corp. 2018, 2022 All Rights Reserved
+# (c) Copyright IBM Corp. 2018, 2023 All Rights Reserved
 # ===========================================================================
 #
 # This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


closed/get_openssl_source.sh
```diff
@@ -72,7 +72,7 @@ case "$OPENSSL_VERSION" in
 		OPENSSL_SOURCE_TAG="openssl-$OPENSSL_VERSION"
 		;;
 	*)
-		usage
+		OPENSSL_SOURCE_TAG=$OPENSSL_VERSION
```
keithc-ca, line 75: While reviewing this, I notice that this script will fail if `OPENSSL_VERSION` includes spaces. I'm not sure whether or not we should be concerned about that. Opinion(s)?

pshipton: I wasn't planning to use any space in a branch or tag name.

keithc-ca: I don't think git allows branch or tag names to have spaces, so you would have no need to specify `OPENSSL_VERSION` to be a value with spaces, but I suppose my question was whether the script should only accept values that are reasonable.

pshipton: We've managed this long without it, it doesn't seem important.

keithc-ca: Agreed, but I thought it better the assumption be mentioned.

keithc-ca:
Please create the changes for newer versions; I'll merge them latest version first.

keithc-ca (approved):

PR #641, opened 2023-04-06
Add additional debug message to show contents of authentication reply message

KostasTsiounis:
Additional information is printed when the DEBUG flag is set for communication with the Kerberos Authentication Server. More specifically, the reply message returned from the Authentication Server is printed in hexadecimal format, to allow for debugging in case of unexpected responses.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/579


keithc-ca:
Same change as in https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/579, tested locally.

PR #647, opened 2023-04-25
Cause a system dump on failure of UncaughtExceptionsTest

pshipton:
Issue https://github.com/eclipse-openj9/openj9/issues/11930


keithc-ca:
Please address the line-endings check failure.

Is this intended to be temporary (until we align openj9 thread names with openjdk)?

keithc-ca (approved):

PR #650, opened 2023-05-03
Add support for EC key generation using native OpenSSL library and restructure ECDH key agreement.

KostasTsiounis:
As part of this update, two things happen:

1. A new class is created and equivalent functionality is added that allows the use the native OpenSSL library to perform EC key generation. As part of that, NativeECKeyPairGenerator.java is created and several other classes are amended to support it, namely SunEC to load the appropriate class, and the ECKeyPairGenerator offered by Sun as a fallback option.
2. The code for the ECDH Key Agreement is updated. In that part, helper functions are created in the NativeCrypto.c file to avoid unnecessary passing of EC_KEY pointers. A NativeECUtil.java class is, also, created, so as to bundle commonly used functionality. Lastly, the handling of freeing arrays, keys and BIGNUM structs that were part of that functionality is updated.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/593

### DIFF
--- a/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
+++ b/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
@@ -48,6 +48,10 @@ public class NativeCrypto {
     public static final int SHA5_384 = 3;
     public static final int SHA5_512 = 4;
 
+    /* Define constants for the EC field types. */
+    public static final int ECField_Fp = 0;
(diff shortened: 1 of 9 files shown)


keithc-ca (approved):

keithc-ca:
Jenkins test sanity win jdk11

knn-k:
I opened https://github.com/eclipse-openj9/openj9/issues/18071, which was caused by this PR.

PR #653, opened 2023-05-05
Get byte array from DerValue without using InputStream

KostasTsiounis:
This fix eliminates the problem of emptying the input stream when reading a DerValue to print for debug purposes.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/597

### DIFF
--- a/src/java.security.jgss/share/classes/sun/security/krb5/internal/KDCRep.java
+++ b/src/java.security.jgss/share/classes/sun/security/krb5/internal/KDCRep.java
@@ -138,20 +138,22 @@ public class KDCRep {
                         " req type is " + req_type);
 
                 System.out.println(">>> KDCRep: Message in bytes is =>");
-                byte[] dataBytes = encoding.getDataBytes();
-                StringBuilder sb = new StringBuilder();
-                for (int i = 0; i < dataBytes.length; i++) {
-                    if ((i % 16) == 0) {
-                        sb.append(String.format("%06X", i));
+                byte[] dataBytes = encoding.getData().toByteArray();
+                if (dataBytes != null) {
+                    StringBuilder sb = new StringBuilder();
+                    for (int i = 0; i < dataBytes.length; i++) {
+                        if ((i % 16) == 0) {
+                            sb.append(String.format("%06X", i));
+                        }
+                        sb.append(String.format(" %02X", dataBytes[i] & 0xFF));
+                        if ((i % 16) == 15) {
+                            System.out.println(sb.toString());
+                            sb.setLength(0);
+                        }
                     }
-                    sb.append(String.format(" %02X", dataBytes[i] & 0xFF));
-                    if ((i % 16) == 15) {
+                    if (sb.length() > 0) {
                         System.out.println(sb.toString());
-                        sb.setLength(0);
                     }
-                }
-                if (sb.length() > 0) {
-                    System.out.println(sb.toString());
                 }
             }
             throw new Asn1Exception(Krb5.ASN1_BAD_ID);


src/java.security.jgss/share/classes/sun/security/krb5/internal/KDCRep.java
```diff
@@ -138,7 +138,7 @@ protected void init(DerValue encoding, int req_type)
                         " req type is " + req_type);
 
                 System.out.println(">>> KDCRep: Message in bytes is =>");
-                byte[] dataBytes = encoding.getDataBytes();
+                byte[] dataBytes = encoding.getData().toByteArray();
```
keithc-ca, line 141: This change must differ from later versions because `toByteArray()` might return `null` (if there are no bytes in encoding). I expect an exception to be thrown by
```java
        der = encoding.getData().getDerValue();
```
on line 159 below, but enabling debugging should replace that with a `NullPointerException` on line 143.

keithc-ca (changes requested):

src/java.security.jgss/share/classes/sun/security/krb5/internal/KDCRep.java
KostasTsiounis: The `toByteArray()` will only throw a `NullPointerException` if the buffer is null. But this cannot happen since the constructor of the original `DerValue` would throw an `IOException`.

Am I missing some path that could cause that `NullPointerException`?

KostasTsiounis: And the constructor of `DerValue` never gets a null argument, since the `KdcComm` class, where the byte array is created, checks and throws an `IOException` itself, if the response byte array is null.

keithc-ca: Notice, I didn't say that `toByteArray()` would throw an exception - I said it might return `null`. See [`DerInputBuffer.toByteArray()`](https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/src/java.base/share/classes/sun/security/util/DerInputBuffer.java#L76).

On line 143, the expression `dataBytes.length` will throw `NullPointerException` if `dataBytes` is `null`. We must not allow that to happen; we need to reach line 157 where the `Asn1Exception` is thrown.

KostasTsiounis: Oh, I see. You mean if the original byte array supplied to the `DerValue` is not null, but is empty, in which case the `DerInputBuffer.toByteArray()` will return null.

So should I just catch the exception and move on, or should I throw the `Asn1Exception` right away?

KostasTsiounis: Changed it to catch the exception and just print another message saying the reply is empty.

src/java.security.jgss/share/classes/sun/security/krb5/internal/KDCRep.java
```diff
@@ -138,20 +138,24 @@ protected void init(DerValue encoding, int req_type)
                         " req type is " + req_type);
 
                 System.out.println(">>> KDCRep: Message in bytes is =>");
-                byte[] dataBytes = encoding.getDataBytes();
-                StringBuilder sb = new StringBuilder();
-                for (int i = 0; i < dataBytes.length; i++) {
-                    if ((i % 16) == 0) {
-                        sb.append(String.format("%06X", i));
+                try {
```
keithc-ca, line 141: This seems inappropriate. Why not simply `if (dataBytes != null) { ... }` (which would also be consistent with later versions).

src/java.security.jgss/share/classes/sun/security/krb5/internal/KDCRep.java
```diff
@@ -138,20 +138,24 @@ protected void init(DerValue encoding, int req_type)
...
                     }
-                    sb.append(String.format(" %02X", dataBytes[i] & 0xFF));
-                    if ((i % 16) == 15) {
-                        System.out.println(sb.toString());
-                        sb.setLength(0);
-                    }
-                }
-                if (sb.length() > 0) {
+                    if (sb.length() > 0) {
                     System.out.println(sb.toString());
```
keithc-ca, line 155: This line is not properly indented.

src/java.security.jgss/share/classes/sun/security/krb5/internal/KDCRep.java
```diff
@@ -138,20 +138,24 @@ protected void init(DerValue encoding, int req_type)
...
+                    }
+                } else {
+                    System.out.println("\tReply message is empty.");
```
keithc-ca, line 158: I don't think this difference (as compared with newer versions) is necessary.

keithc-ca (changes requested):

src/java.security.jgss/share/classes/sun/security/krb5/internal/KDCRep.java
KostasTsiounis: Done.

src/java.security.jgss/share/classes/sun/security/krb5/internal/KDCRep.java
keithc-ca: This was not corrected in the latest commit.

KostasTsiounis: Sorry about that. It was looking OK on my IDE and didn't check on Github. I fixed it.

keithc-ca:

keithc-ca (approved):

KostasTsiounis:
FYI @keithc-ca

PR #652, opened 2023-05-04
Merge master jdk-11.0.20+1 into openj9-staging

JasonFengJ9:
Resolved conflicts:
```
	.gitignore
```

depends 
* https://github.com/eclipse-openj9/openj9/pull/17339

Signed-off-by: Jason Feng <fengj@ca.ibm.com>

PR #654, opened 2023-05-08
Remove useless patterns from .gitignore

keithc-ca:
Noticed while reviewing #652 (see https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/652#issuecomment-1534893774).

* configure option `--with-freetype-src` has been removed, so `freetype.bat` and `freetype.log` should not be created
* I don't think `/.build` ever needed to be ignored


JasonFengJ9 (approved):
LGTM 

No PR build required

PR #655, opened 2023-05-11
Cause a system dump on failure of UncaughtExceptionsTest

pshipton:
Seems I caught the exception on the wrong line. Include all the cases.

Related to https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/647 and https://github.com/eclipse-openj9/openj9/issues/11930#issuecomment-1544458452


keithc-ca (approved):
Is the intent to apply this change to newer versions as well, or just to Java 11 to help diagnose the underlying problem?

pshipton:
Just jdk11. I'm only seeing the failure on jdk11. we excluded the test on jdk17 and I expect that's carried forward. If we can fix the problem we can unexclude the test in all versions.

PR #659, opened 2023-05-23
Fix mode of ReleasePrimitiveArrayCritical to ensure copying of data

KostasTsiounis:
Change mode value from `JNI_ABORT` to `0` in `ReleasePrimitiveArrayCritical` of arrays pertaining to EC and PBE key generation.

This makes no difference if a direct reference to the array is provided by the JVM, but it ensures copying of data to the original array, in case a copy of it was provided when calling `GetPrimitiveArrayCritical`.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/603


keithc-ca (approved):

keithc-ca:
Jenkins test sanity win jdk11

PR #672, opened 2023-06-21
Enable CRIU support builds for AArch64

0xdaryl:


tajila:
jenkins test sanity alinux64 jdk11

closed/autoconf/custom-hook.m4
```diff
@@ -377,7 +377,7 @@ AC_DEFUN([OPENJ9_CONFIGURE_CRIU_SUPPORT],
     AC_MSG_RESULT([no (explicitly disabled)])
   elif test "x$enable_criu_support" = x ; then
     case "$OPENJ9_PLATFORM_CODE" in
-      xa64|xz64)
+      xa64|xz64|xr64)
```
keithc-ca, line 380: We aim to keep lists like this sorted (when it otherwise doesn't matter) making it easier for humans to comprehend.

PR #675, opened 2023-06-22
Add support for a suspendOnRestore option in JDWP

mikezhang1234567890:
To help debug issues with checkpoint/restore, add an option similar to "suspend=y" which suspends on VM startup except this new option will suspend when the VM is restored from a checkpoint.
To that end, this PR adds support for a VM_Restore JVMTI event to JDWP and JDB.

Port of https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/591 to JDK11.


JasonFengJ9:
Superseded by https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/803.

PR #676, opened 2023-06-27
Improve speed of cipher operations for large inputs and allow unpadding at offset of output

KostasTsiounis:
In cases were bytes remain buffered and/or padding is enabled and encryption is performed, the entirety of the input provided in the `doFinal()` methods of the `CipherCore` class would be copied in a buffer with enough space to accommodate the extra bytes requires for buffered bytes and/or padding. The buffer would subsequently be zeroed out for security reasons.

To avoid said actions and improve the performance of the process in case of very large inputs, the updated code calls the `update()` method with as much data as possible before the final steps.

Changes made as part of https://github.com/ibmruntimes/openj9-openjdk-jdk17/commit/d8fc6df82151f89566508fcb9da1e18f9a236527 had to, also, be back-ported to allow unpadding to happen at a specific offset of the output. Those involved adding an extra parameter in the `unpad()` method and using it when invoking the corresponding method of the assigned `Padding`.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/614

### DIFF
--- a/src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
+++ b/src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
@@ -897,12 +897,26 @@ final class CipherCore {
         try {
             checkReinit();
             byte[] output = new byte[getOutputSizeByOperation(inputLen, true)];
-            byte[] finalBuf = prepareInputBuffer(input, inputOffset,
-                    inputLen, output, 0);
+            int outputOffset = 0;
+            int outLen = 0;
+
+            if (inputLen > 0) {
+                /*
+                 * Call the update() method to get rid of as many bytes as
+                 * possible before potentially copying array.
+                 */
+                int updateOutLen = update(input, inputOffset, inputLen, output, outputOffset);
+                inputOffset = inputLen;
+                inputLen = 0;
+                outputOffset += updateOutLen;
+                outLen = updateOutLen;
+            }
+
+            byte[] finalBuf = prepareInputBuffer(input, inputOffset, inputLen, output, outputOffset);
             int finalOffset = (finalBuf == input) ? inputOffset : 0;
             int finalBufLen = (finalBuf == input) ? inputLen : finalBuf.length;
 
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/com/sun/crypto/provider/CipherCore.java
```diff
@@ -1026,16 +1058,15 @@ private void endDoFinal() {
         }
     }
 
-    private int unpad(int outLen, byte[] outWithPadding)
+    private int unpad(int outLen, int off, byte[] outWithPadding)
```
keithc-ca, line 1061: This appears to be a back-port of part of
* [8253821: Improve ByteBuffer performance with GCM](https://github.com/ibmruntimes/openj9-openjdk-jdk11/commit/d8fc6df82151f89566508fcb9da1e18f9a236527)

@KostasTsiounis Is that your understanding?

That seems reasonable, but should have been called out in the commit message and the description here.

KostasTsiounis: I didn't see this PR you are mentioning. I needed the extra functionality for my change to work and it wasn't available in OpenJDK11.

Should I change the description and/or the commit message. And if yes, what would the change be?

keithc-ca: The commit I mentioned above is derived from upstream; see also https://bugs.openjdk.org/browse/JDK-8253821 which says it was applied to Java 16. The commit message and the description should discuss that.

pshipton: In our jdk17 repo we have https://github.com/ibmruntimes/openj9-openjdk-jdk17/commit/d8fc6df82151

keithc-ca: The commit message is a little terse, but there's enough information here now.

keithc-ca (approved):

keithc-ca:
Tested internally.

PR #687, opened 2023-08-01
Enable JITServer build by default on AArch64 Linux

knn-k:
This commit enables JITServer build by default on AArch64 Linux (xr64).


keithc-ca:
Jenkins test sanity alinux64 jdk11

keithc-ca (approved):

PR #692, opened 2023-08-10
Fix harfbuzz for AIX

keithc-ca:
The update to harfbuzz 7.2.0 together with [8307603: [AIX] Broken build after JDK-8307301](https://github.com/ibmruntimes/openj9-openjdk-jdk11/commit/f8aa576a5e31c877edbfe9bafd30c08904b1f3d4)
will cause the problem discussed in https://github.com/eclipse-openj9/openj9/issues/17759 to return. This should fix that.

Note this targets the openj9-staging branch.

### DIFF
--- a/src/java.base/share/classes/java/util/Timer.java
+++ b/src/java.base/share/classes/java/util/Timer.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2022, 2022 All Rights Reserved
+ * (c) Copyright IBM Corp. 2022, 2023 All Rights Reserved
  * ===========================================================================
  */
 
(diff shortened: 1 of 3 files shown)


keithc-ca:
Jenkins compile aix jdk11

pshipton (approved):

pshipton:
@backwaterred FYI. While removing hb-ot-post-table-v2subset.h via [8307603: [AIX] Broken build after JDK-8307301](https://github.com/ibmruntimes/openj9-openjdk-jdk11/commit/f8aa576a5e31c877edbfe9bafd30c08904b1f3d4) allows the code to compile on AIX, the library can't be loaded at runtime due to missing symbols, see https://github.com/eclipse-openj9/openj9/issues/17759. Keith's backport from the Harfbuzz project is a better fix. There is a little more info about [this change](https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/692/commits/82e2ec753bf09514d532590ac084211e99ff69d5) in https://github.com/ibmruntimes/openj9-openjdk-jdk17/pull/235. It seems the problem will be resolved once harfbuzz is updated to 8.x.

PR #701, opened 2023-09-06
Cache openssl message digest contexts

jasonkatonica:
It has been identified that when using `OpenSSL` version `3.x` the message digest context initialization is much more expensive than using `OpenSSL` version `1.x`. This update optimizes the use of `OpenSSL` contexts by making use of a simple cache.

We will create a second initialized openssl digest context when the message digest cipher is initialized. This second context, defined as `context->cachedInitalizedDigestContext`, will be copied over the working context, `context->ctx`, using the `EVP_MD_CTX_copy_ex` API whenever we wish to re-initalize this cipher. The restoration of a context occurs whenever a cipher is explictly reset or whenever a final digest is computed.

Backported from:

https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/643

Signed-off by: Jason Katonica <katonica@us.ibm.com>

### DIFF
--- a/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
+++ b/closed/src/java.base/share/classes/jdk/crypto/jniprovider/NativeCrypto.java
@@ -218,7 +218,7 @@ public class NativeCrypto {
                                                   int digestOffset,
                                                   int digestLen);
 
-    public final native void DigestReset(long context);
+    public final native int DigestReset(long context);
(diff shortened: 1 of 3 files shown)


keithc-ca:
Jenkins compile osx jdk11

keithc-ca (approved):
Testing passed in https://openj9-jenkins.osuosl.org/job/Grinder/2879.

PR #707, opened 2023-09-13
Modify get_source.sh to automatically fetch release branches

pshipton:
If the current branch matches the pattern of a release branch, fetch openj9/omr branches of the same name.

This will work for a normal release, and can be overwritten for special cases. It avoids having to modify a release branch after creation in order to get the matching openj9/omr release branches. Further changes to the release branch can be pushed rather than merged, as long as the release branch matches the head stream.

Backport of https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/660

### DIFF
--- a/closed/get_j9_source.sh
+++ b/closed/get_j9_source.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # ===========================================================================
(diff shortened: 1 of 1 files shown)


keithc-ca (approved):

keithc-ca:
Jenkins compile win jdk11

PR #708, opened 2023-09-13
Merge latest openj9 changes to 0.41

pshipton:
Manual merge, fixing the merge conflict in get_j9_source.sh

### DIFF
--- a/closed/autoconf/custom-hook.m4
+++ b/closed/autoconf/custom-hook.m4
@@ -377,7 +377,7 @@ AC_DEFUN([OPENJ9_CONFIGURE_CRIU_SUPPORT],
     AC_MSG_RESULT([no (explicitly disabled)])
   elif test "x$enable_criu_support" = x ; then
     case "$OPENJ9_PLATFORM_CODE" in
-      xa64|xr64|xz64)
+      xa64|xl64|xr64|xz64)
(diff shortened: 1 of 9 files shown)


JasonFengJ9 (approved):

keithc-ca:
This doesn't appear to have achieved what I think you expected for `get_j9_source.sh`; see https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/v0.41.0-release/closed/get_j9_source.sh#L59-L63.

pshipton:
I did that intentionally, to avoid any accidental breakage from a new largely untested change. 0.41 has already diverged from the head stream so there is no benefit. We'll use the change for 0.42 to avoid having to create PRs like this one. I may end up always hard coding the branch names before we ship to avoid side effects.

keithc-ca:
> I may end up always hard coding the branch names

Wasn't the whole point of that series of pull requests to avoid the need for those manual changes?

pshipton:
No, it's not the manual change I wanted to avoid, but having to create PRs like this to pick up the latest changes.

keithc-ca:
> If we continue to make the manual change

My preference is that we don't do that.

pshipton:
Sure, we can see how it goes, and if anyone complains about the magic.

PR #714, opened 2023-09-21
Look for -XX:[+/-]UseZlibNX on AIX

pshipton:
AIX puts zlibNX on the LIBPATH when running on P9 or later, but having this in the environment may cause some things, such as git clone, to fail when exec'ed from Java. For example,
`fatal: pack has bad object at offset 3872100: inflate returned -5`

Backport of https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/664


src/java.base/unix/native/libjli/java_md_solinux.c
```diff
@@ -402,7 +425,7 @@ CreateExecutionEnvironment(int *pargc, char ***pargv,
                     /* On AIX we additionally need 'jli' in the path because ld doesn't support $ORIGIN. */
                     JLI_StrLen(jrepath) + JLI_StrLen("/lib//jli:") +
                     /* On AIX P9 or newer with NX accelerator enabled, add the accelerated zlibNX to LIBPATH */
-                    ((power_9_andup() && power_nx_gzip()) ? JLI_StrLen(":" ZLIBNX_PATH) : 0) +
```
keithc-ca, line 405: While you're here, can you remove the extra `/` on line 426? It's leftover from changes in jdk8 where there's the extra "arch" directory.

pshipton: Updated.

keithc-ca:
Jenkins compile aix jdk11

PR #717, opened 2023-10-02
Always revert to Java impl when OpenSSL fails

KostasTsiounis:
At the moment, if there is an OpenSSL failure when using `NativeECKeyPairGenerator`, a flag for that particular curve is set and the Sun Java implementation is used to complete the task.

If another instance of `NativeECKeyPairGenerator` reaches the same failure, a `ProviderException` is thrown, as the issue with that particular curve should have been discovered during the call to `initialize()` thus creating and using an instance of `ECKeyPairGenerator`, and that particular point should never have been reached.

However, in the scenario where the second instance of `NativeECKeyPairGenerator` has been initialized earlier, the flag for the problematic curve will never be discovered.

This fix ensures that we always revert to the Sun Java implementation when an OpenSSL failure occurs.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/672


keithc-ca:
Jenkins compile zlinux jdk11

keithc-ca (approved):

PR #719, opened 2023-10-10
Remove the visitAnnotation for Hidden in generateInvokerTemplate

ThanHenderson:
This patch fixes both eclipse-openj9/openj9#14553 and eclipse-openj9/openj9#18245. The visitAnnotation for Hidden seems to have unwanted side effects when OJDK MHs are enabled and invokeWithArguments is used in JDK11. The behaviour when removing the visitAnnotation matches the behaviour of the RI and the implementation for JDK17.

Closes: eclipse-openj9/openj9#14553 eclipse-openj9/openj9#18245
Signed-off-by: Nathan Henderson <nathan.henderson@ibm.com>

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/MethodHandleImpl.java
+++ b/src/java.base/share/classes/java/lang/invoke/MethodHandleImpl.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2022, 2022 All Rights Reserved
+ * (c) Copyright IBM Corp. 2022, 2023 All Rights Reserved
  * ===========================================================================
  */
 
@@ -33,7 +33,6 @@ package java.lang.invoke;
 
 import jdk.internal.misc.JavaLangInvokeAccess;
 import jdk.internal.misc.SharedSecrets;
-import jdk.internal.org.objectweb.asm.AnnotationVisitor;
 import jdk.internal.org.objectweb.asm.ClassWriter;
 import jdk.internal.org.objectweb.asm.MethodVisitor;
 import jdk.internal.reflect.CallerSensitive;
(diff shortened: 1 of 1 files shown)


ThanHenderson:
Jenkins without patch: https://hyc-runtimes-jenkins.swg-devops.com/job/Pipeline-Build-Test-Personal/18477/

Jenkins with patch: https://hyc-runtimes-jenkins.swg-devops.com/job/Pipeline-Build-Test-Personal/18480/

@babsingh

babsingh:
In JDK17, the annotation was removed due to the move towards hidden classes: https://github.com/ibmruntimes/openj9-openjdk-jdk17/commit/24200973d4827635b481415ce5e65573b356400e#diff-4e4fca8bb2eb6320ff485ee724248e1641b4bb3f6dbae8526e87c5cf15905d9a.

In JDK11, hidden classes are not supported. So, removing the hidden annotation will show the impacted method in the stack trace. We won't match the RI in terms of what is shown in the stack trace. Also, other methods which rely upon the stack trace might stop working.

https://github.com/eclipse-openj9/openj9/issues/14553: The core issue is `Caused by: java.lang.InternalError: found java.lang.invoke.MethodHandleImpl$BindCaller, expected InjectedInvoker/0x0000000000000000`. 

https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/b40478156401262ba42d8dc5f7c838c03df5ce5d/src/java.base/share/classes/java/lang/invoke/MethodHandleImpl.java#L1300

```
        @CallerSensitive
        @ForceInline // to ensure Reflection.getCallerClass optimization
        private static boolean checkCallerClass(Class<?> expected) {
            // This method is called via MH_checkCallerClass and so it's correct to ask for the immediate caller here.
            Class<?> actual = Reflection.getCallerClass();
            if (actual != expected)
                throw new InternalError("found " + actual.getName() + ", expected " + expected.getName()); <--- ERROR
            return true;
        }
```

Do we know why the `actual` class doesn't match the `expected` class? `Reflection.getCallerClass` retrieves the `actual` class ... where/how the value of the `actual` class is retrieved and set?

PR #737, opened 2023-12-06
Enable ECDSA ciphers when running with FIPS cryptography

jasonkatonica:
In FIPS mode, certain cryptographic algorithms are typically disabled. This includes the algorithm `SHA1withECDSA` since `SHA1` is no longer recommended for general use.

The `JSSE`provider, responsible for managing TLS cipher specifications, currently disables all `*ECDSA*` ciphers due to the absence of the `SHA1withECDSA` algorithm on the system when running in FIPS mode. This behavior is inaccurate, since other ECDSA-related transformations, such as `SHA512withECDSA` or `SHA384withECDSA`, are available in a FIPS environment and can be used for TLS connections.

With this update, `*ECDSA*` ciphers can now be enabled in FIPS mode. We will allow for `*ECDSA*` ciphers to be enabled if any algorithm in the ECDSA family is present such as `SHA512withECDSA`, `SHA384withECDSA`, `SHA256withECDSA`, or `SHA224withECDSA`. We expect all these algorithms to be present in the various FIPS solutions available within Semeru.

### DIFF
--- a/src/java.base/share/classes/sun/security/ssl/JsseJce.java
+++ b/src/java.base/share/classes/sun/security/ssl/JsseJce.java
@@ -23,6 +23,12 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2023, 2023 All Rights Reserved
+ * ===========================================================================
+ */
+
 package sun.security.ssl;
 
 import java.math.BigInteger;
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/sun/security/ssl/JsseJce.java
```diff
@@ -393,10 +419,38 @@ private static class EcAvailability {
...
+        /**
+         * Checks if a particular signature algorithm is available.
+         *
+         * @param algorithm the algorithm we will attempt to instantiate to check if it is available
+         * @return true if the signature algorithm is found, false otherwise
+         */
+        private static boolean isSignatureAlgorithmAvailable(String algorithm) {
+            try {
+                // Attempt to create a Cipher instance with the specified algorithm.
+                Signature.getInstance(algorithm);
```
keithc-ca, line 431: I think this should use `JsseJce.getSignature()` like line 399 did.

keithc-ca (changes requested):

src/java.base/share/classes/sun/security/ssl/JsseJce.java
jasonkatonica: Updated to make use of `JsseJce.getSignature()` instead of `Signature.getInstance()`

keithc-ca:
Jenkins compile plinux jdk11

PR #744, opened 2023-12-12
Revert to Java impl when non-default SecureRandom present

KostasTsiounis:
If one initializes any of the `NativeECKeyPairGenerator` or `NativeXDHKeyPairGenerator` with a `SecureRandom` implementation, it is effectively ignored, as the equivalent OpenSSL code is used instead.

This fix changes that to using OpenSSL only when no or the default Sun implementation of `SecureRandom` is provided. In any other case, the original Java implementation is used to generate the appropriate `KeyPair`.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/711

Signed-off by: Kostas Tsiounis kostas.tsiounis@ibm.com

### DIFF
--- a/src/java.base/share/classes/module-info.java
+++ b/src/java.base/share/classes/module-info.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2022, 2022 All Rights Reserved
+ * (c) Copyright IBM Corp. 2022, 2023 All Rights Reserved
  * ===========================================================================
  */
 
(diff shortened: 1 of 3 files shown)


src/java.base/share/classes/module-info.java
```diff
@@ -298,6 +298,7 @@
         java.rmi,
         java.security.jgss,
         jdk.crypto.cryptoki,
+        jdk.crypto.ec,
```
pshipton, line 301: Pls update the IBM copyright end date in this file.

KostasTsiounis: Done.

pshipton:
jenkins copyright check

pshipton:
jenkins compile amac jdk11

pshipton (approved):

PR #746, opened 2024-01-04
Restore behavior on platforms where CRIU is supported but not in use

WilburZjh:
This PR aims to fix an issue on how to read random data and
get random bytes on platforms where CRIU is supported, but
not in use by enabling it via a --enable-criu-support JVM
option.

Issue https://github.com/eclipse-openj9/openj9/issues/18602


WilburZjh:
Build is passed, and the failed cases are passed. Change from draft to ready for review.

@jasonkatonica FYI

pshipton (approved):

PR #748, opened 2024-01-09
Use correct JNI function to call ArrayList.add()

jasonkatonica:
Various testcases are being run as part of the `openj9` project. Currently one of these tests is failing since they are able to detect that the wrong method signature is being used by JNI calls at runtime.

This update makes the same method call only it expects a boolean to be returned from the method call. This allows for the test to pass the JNI method signature check.

This fixes [openj9 issue 17795](https://github.com/eclipse-openj9/openj9/issues/17795).


keithc-ca (approved):

keithc-ca:
Jenkins compile win jdk11

PR #751, opened 2024-01-15
Add provider name and class name mapping in Restricted Security mode

taoliult:
This is a back-port PR from PR https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/722

For the Java security providers, for example, the SunSASL, its class name is com.sun.security.sasl.Provider. From the provider class name, can not get the provider name which defined in its construction method. So, add the mapping between the provider name and its class name in Restricted Security mode.


keithc-ca (approved):

keithc-ca:
Jenkins test sanity xlinux jdk11

PR #757, opened 2024-02-08
Disable CRAC_SUPPORT by default for all platforms

JasonFengJ9:
Disable `CRAC_SUPPORT` by default for all platforms

Cherry-pick
* https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/743

Depended on
* https://github.com/adoptium/TKG/pull/497
* https://github.com/eclipse-openj9/openj9/pull/18928

This fixes an internal RTC workitem 150440.


JasonFengJ9:
This isn't needed since `exports jdk.crac;` will be removed by an OpenJ9 PR (still in progress), and the package `jdk.crac` will be exported programmatically at runtime when the CRaC API is in use, i.e., `-XX:CRaCCheckpointTo` is specified.

PR #759, opened 2024-02-29
Use Unsafe.compareAndExchange for writing to Stable field and array

nbhuiyan:
Using Unsafe.compareAndExchange prevents issues resulting from multiple threads coming across uninitialized VarHandle's typesAndInvokers field and competing to populate the field with their own versions.

Using compareAndExchange to populate the typesAndInvokers' MethodHandle table also prevents similar issues when multiple threads come across the same unpopulated slot in the MethodHandle table and compete to populate it.

Issue: eclipse-openj9/openj9#18882

~In Draft state until personal builds/tests pass.~

### DIFF
--- a/src/java.base/share/classes/java/lang/invoke/VarHandle.java
+++ b/src/java.base/share/classes/java/lang/invoke/VarHandle.java
@@ -23,9 +23,16 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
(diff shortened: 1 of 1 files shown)


nbhuiyan:
@keithc-ca Ready for review.

keithc-ca (approved):

keithc-ca:
Verified locally in a build configured with `--enable-openjdk-methodhandles`.

PR #761, opened 2024-03-14
Integrate OpenJCEPlus into Semeru OpenJDK

JinhangZhang:
Integrate the building of OpenJCEPlus with Semeru
OpenJDK as a module "openjceplus". The java codes
will be built with Semeru OpenJDK as a jmod, and
the native codes will be built as a ".so" or ".dll" library.

### DIFF
--- a/get_source.sh
+++ b/get_source.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 # ===========================================================================
-# (c) Copyright IBM Corp. 2017, 2022 All Rights Reserved
+# (c) Copyright IBM Corp. 2017, 2024 All Rights Reserved
 # ===========================================================================
 #
 # This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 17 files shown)


JinhangZhang:
@jasonkatonica FYI

get_source.sh
```diff
@@ -57,7 +64,23 @@ for i in "$@" ; do
 			usage
 			;;
 
-		-openj9-repo=* | -openj9-branch=* | -openj9-sha=* | -openj9-reference=* | -omr-repo=* | -omr-branch=* | -omr-sha=* | -omr-reference=* | -parallel=* )
+		-gskit-bin=* \
```
keithc-ca, line 67: Please indent this the same as in later versions (two more spaces).

src/java.base/share/classes/module-info.java
```diff
@@ -160,6 +160,9 @@
         jdk.compiler,
         jdk.jlink;
     exports jdk.internal.logger to
+    /*[IF OPENJCEPLUS_SUPPORT]*/
+        openjceplus,
+    /*[ENDIF] OPENJCEPLUS_SUPPORT */
```
keithc-ca, line 165: Please don't indent the preprocessor comment, for consistency with later versions:
```java
    exports jdk.internal.logger to
/*[IF OPENJCEPLUS_SUPPORT]*/
        openjceplus,
/*[ENDIF] OPENJCEPLUS_SUPPORT */
```

keithc-ca (changes requested):

src/java.base/share/classes/module-info.java
JinhangZhang: updated

get_source.sh
```diff
@@ -57,7 +64,23 @@ for i in "$@" ; do
 			usage
 			;;
 
-		-openj9-repo=* | -openj9-branch=* | -openj9-sha=* | -openj9-reference=* | -omr-repo=* | -omr-branch=* | -omr-sha=* | -omr-reference=* | -parallel=* )
+	      -gskit-bin=* \
```
keithc-ca, line 67: This line should be identical to the [corresponding line in the head stream](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/get_source.sh#L67).

JinhangZhang: updated. (It seems to be my IDE problem)

keithc-ca (approved):

PR #762, opened 2024-03-18
Parse jar index when using SCC

hangshao0:
If a jarIndex exists, the JDK by default (i.e. SCC is off) uses the URLs in the JarIndex to search for resources, the class path in manifest is ignored.

However, when SCC is on, the jarIndex is always ignored and the class path in manifest is used, which differs from the default JDK behavior.

Change SCC code in URLClassPath to always use URLs in jarIndex if it exists.

Add a property "com.ibm.oti.shared.disableJarIndex" that can be used to control the behavior.

JarIndex is removed in Java 21, no need to port to Java 21+.

Port of https://github.com/ibmruntimes/openj9-openjdk-jdk17/pull/331/

Internal issue semeru-runtimes/issues/1

### DIFF
--- a/src/java.base/share/classes/jdk/internal/loader/URLClassPath.java
+++ b/src/java.base/share/classes/jdk/internal/loader/URLClassPath.java
@@ -24,7 +24,7 @@
  */
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 1997, 2020 All Rights Reserved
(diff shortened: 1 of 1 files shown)


pshipton:
jenkins test sanity,extended amac jdk11

pshipton:
The CryptoTests_0 failure is unrelated.

pshipton (approved):

PR #769, opened 2024-04-17
Check setting of security property when in FIPS mode

KostasTsiounis:
When in `FIPS` mode, calling `Security.setProperty()` should not be allowed when the security property that the user is trying to override is set in the active `RestrictedSecurity` profile. A `SecurityException` is thrown if such an attempt is made.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/773


keithc-ca:
Jenkins test sanity plinux jdk11

keithc-ca (approved):

PR #779, opened 2024-05-08
Remove the 192-bit EC test case

JinhangZhang:
This patch eliminates the 192-bit EC test which causes exceptions
seen in issue #18320.

DefaultSignatureAlgorithm test was run in those Redhat OS based machines
in a non-FIPS mode, but with a FIPS version of openssl. So, a 192-bit 
size of EC key pair generator is not allowed by the native code in a
FIPS version of openssl. The code path went to a replacement EC key-pair
generator Java implementation.

issue: https://github.com/eclipse-openj9/openj9/issues/18320

### DIFF
--- a/test/jdk/sun/security/tools/keytool/fakegen/DefaultSignatureAlgorithm.java
+++ b/test/jdk/sun/security/tools/keytool/fakegen/DefaultSignatureAlgorithm.java
@@ -60,7 +60,6 @@ public class DefaultSignatureAlgorithm {
         check("DSA", 1024, null, "SHA256withDSA");
         check("DSA", 3072, null, "SHA256withDSA");
 
-        check("EC", 192, null, "SHA256withECDSA");
         check("EC", 384, null, "SHA384withECDSA");
         check("EC", 571, null, "SHA512withECDSA");
 


JinhangZhang:
@jasonkatonica @pshipton  FYI

pshipton:
If you are referring to an issue pls provide a link not just a number.
This should be added to the head stream (openj9-openjdk-jdk) first and then backported. If it doesn't apply to newer versions, pls explain.

JinhangZhang:
> If you are referring to an issue pls provide a link not just a number. This should be added to the head stream (openj9-openjdk-jdk) first and then backported. If it doesn't apply to newer versions, pls explain.

@pshipton This only happens at 11. Case [18320](https://github.com/eclipse-openj9/openj9/issues/18320) is failed on this [line](https://github.com/ibmruntimes/openj9-openjdk-jdk11/blob/openj9/test/jdk/sun/security/tools/keytool/fakegen/DefaultSignatureAlgorithm.java#L63) of code. The size of 192 is not supported by a FIPS version of OpenSSL. This 192 size of EC key is not declared in other versions such as [17](https://github.com/ibmruntimes/openj9-openjdk-jdk17/blob/openj9/test/jdk/sun/security/tools/keytool/fakegen/DefaultSignatureAlgorithm.java), [21](https://github.com/JinhangZhang/openj9-openjdk-jdk21/blob/openj9/test/jdk/sun/security/tools/keytool/fakegen/DefaultSignatureAlgorithm.java) and [next](https://github.com/ibmruntimes/openj9-openjdk-jdk/blob/openj9/test/jdk/sun/security/tools/keytool/fakegen/DefaultSignatureAlgorithm.java).

keithc-ca:
Perhaps it makes more sense to just remove the 192-bit test from `DefaultSignatureAlgorithm` (or at least in FIPS mode)?

JinhangZhang:
> Perhaps it makes more sense to just remove the 192-bit test from `DefaultSignatureAlgorithm` (or at least in FIPS mode)?

Remove the 192-bit test to align with other versions makes sense to me.

JinhangZhang:
Removed 192-bit test

pshipton:
Isn't that just leaving the problem to be found by a customer?

keithc-ca:
Please update the commit message and the description here to more accurately describe this change.

JinhangZhang:
> Please update the commit message and the description here to more accurately describe this change.

updated

JinhangZhang:
> > > Please update the commit message and the description here to more accurately describe this change.
> > 
> > 
> > updated
> 
> This doesn't "implement" anything as the commit message and description say, instead it removes testing of a specific EC key size. Perhaps the summary should be (paraphrasing the last line and editing for grammar):
> 
> > Remove test of 192-bit EC or consistency with other JDK versions

Updated

keithc-ca:
All plinux build systems are offline.

keithc-ca:
Jenkins compile zlinux jdk11

keithc-ca (approved):

PR #788, opened 2024-06-20
Allow extension of RestrictedSecurity profiles

KostasTsiounis:
`RestrictedSecurity` profiles sometimes share a lot of duplicate settings with only minor differences. With these changes the extension, similar to object-orientation, of profiles becomes possible.

More specifically, a profile `A` can extend another a profile `B`, using `RestrictedSecurity.<profile A name>.extends = RestrictedSecurity.<profile B name>`. This allows profile `A` to inherit all of profile `B`'s properties. One can add additional properties to profile `A`, or amend some of the existing ones. That includes overriding, appending or removing from a property (wherever that's applicable).

An additional property is introduced. The `RestrictedSecurity.<profile name>.desc.hash = <hash algorithm>:<hash>` is used to ensure the profile hasn't been unintentionally altered. The profile's properties are hashed using the selected `<hash algorithm>`, and the result is compared to the `<hash>` provided through the property. This property is mandatory for base profiles (i.e., profiles that are not extending anything), and optional for the rest.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/793

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -23,7 +23,10 @@
  */
 package openj9.internal.security;
 
+import java.nio.charset.StandardCharsets;
 import java.security.AccessController;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.security.PrivilegedAction;
 import java.security.Provider.Service;
 import java.time.LocalDate;
@@ -42,6 +45,8 @@ import java.util.Properties;
 import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import sun.security.util.Debug;
 
@@ -70,6 +75,8 @@ public final class RestrictedSecurity {
     private static boolean securityEnabled;
 
     private static String userSecurityID;
+
+    private static ProfileParser profileParser;
 
     private static RestrictedSecurityProperties restricts;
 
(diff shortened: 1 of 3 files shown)


KostasTsiounis:
FYI @keithc-ca

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -948,34 +774,52 @@ boolean isRestrictedServiceAllowed(Service service) {
                 return false;
             } else if (constraints.length == 0) {
                 // Allow this provider with no constraints.
+                if (debug != null) {
```
keithc-ca, line 777: Similar to https://github.com/ibmruntimes/openj9-openjdk-jdk17/pull/359#pullrequestreview-2131242787, this file should not have unnecessary differences relative to later versions. Please add the debug code before line 774, the blank line before line 870 and update the comment on line 892.

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Done.

keithc-ca:
Jenkins test sanity xlinux jdk11

keithc-ca:
Jenkins test sanity xlinux jdk11

keithc-ca:
Restarting the part of testing that failed to fetch from artifactory...

keithc-ca:
The rest of the tests passed on a different system in https://openj9-jenkins.osuosl.org/job/Test_openjdk11_j9_sanity.functional_x86-64_linux_Personal_testList_1/341.

keithc-ca (approved):

PR #791, opened 2024-06-26
wip Stable and unique Lambda and LambdaForm class names

ThanHenderson:
This WIP patch replaces the index number in `Lambda` class names with a stable hash-code, and adds a stable hash code to customized `LambdaForm` classes.

The intent is to enable:
1. `LambdaForms` to be stored in SCC
2. early lookup in the SCC for both `Lambda` and `LambdaForm` classes
3. reduce or eliminate the number of class byte comparisons for `Lambdas` when creating ROM classes

Signed-off-by: Nathan Henderson <nathan.henderson@ibm.com>


ThanHenderson:
@babsingh @hangshao0 Could you guys take a look?

My main point of concern here is "uniqueness" of the `uniqueID`s. The `Lambda` and `LambdaForm` class names are stable with these `uniqueID`s included, and the probability of collisions is low but not impossible. This would cause issues if we are doing an early lookup to produce a class from the SCC based only on the class name. In practice, I haven't observed any collisions.

I am open to any other suggestions in how we should be generating the `uniqueID`.

PR #792, opened 2024-06-28
Back-port changes to .copyrightignore from the head stream

keithc-ca:


pshipton (approved):

.copyrightignore
```diff
@@ -20,31 +20,32 @@
...
+*.ico
+*.icu
 *.ini
-.copyrightignore
 *.jpg
-*.ico
-*.bmp
-*.wav
-/*LICENSE
-/*README
```
pshipton, line 48: FYI LICENSE was added to all the versions, but not the README.md

pshipton (approved):

.copyrightignore
keithc-ca: There is no `README`, but `README.md` is covered by `*.md`.

PR #789, opened 2024-06-25
Support provider fully-qualified class name in Restricted Security mode

taoliult:
This is a back port PR from JDKNext PR https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/758

This PR is for updating the codes to only support the fully-qualified class name when setting the providers list in restricted security profiles. For example, the NSS FIPS140-2 provider list will be:

```
RestrictedSecurity.NSS.140-2.jce.provider.1 = sun.security.pkcs11.SunPKCS11 ${java.home}/conf/security/nss.fips.cfg
RestrictedSecurity.NSS.140-2.jce.provider.2 = sun.security.provider.Sun [ \
    {CertificateFactory, X.509, ImplementedIn=Software}, \
    {CertStore, Collection, ImplementedIn=Software}, \
    {CertStore, com.sun.security.IndexedCollection, ImplementedIn=Software}, \
    {Policy, JavaPolicy, *}, {Configuration, JavaLoginConfig, *}, \
    {CertPathBuilder, PKIX, ValidationAlgorithm=RFC5280:ImplementedIn=Software}, \
    {CertPathValidator, PKIX, ValidationAlgorithm=RFC5280:ImplementedIn=Software}, \
    {KeyStore, PKCS12, *}]
RestrictedSecurity.NSS.140-2.jce.provider.3 = sun.security.ec.SunEC [{KeyFactory, EC, ImplementedIn=Software: \
    SupportedKeyClasses=java.security.interfaces.ECPublicKey|java.security.interfaces.ECPrivateKey: \
    KeySize=256}, {AlgorithmParameters, EC, *}]
RestrictedSecurity.NSS.140-2.jce.provider.4 = sun.security.ssl.SunJSSE
```

PR #793, opened 2024-07-04
Update SunJSSE fully-qualified class name in Restricted Security mode

taoliult:
Found a bug in my back-ported JDK11 PR https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/789.

In JDK8 or JDK11, the fully-qualified class name for SunJSSE is `com.sun.net.ssl.internal.ssl.Provider`, not `sun.security.ssl.SunJSSE`. From JDK17, its fully-qualified class name changed to `sun.security.ssl.SunJSSE`.

So, create this PR to update the SunJSSE fully-qualified class name only for JDK11.


keithc-ca:

keithc-ca (approved):

pshipton:
The change is added to the 0.46 release.

PR #795, opened 2024-07-25
Update ignore list based on M2 build

JinhangZhang:


JinhangZhang:
@jasonkatonica @llxia FYI

PR #801, opened 2024-08-09
Disable more DHE related ciphersuites

JinhangZhang:
DHE related cipher suites need Diffie-Hellman crypto services. However, those crypto services are not
allowed in strict profile in FIPS140-3.

### DIFF
--- a/src/java.base/share/conf/security/java.security
+++ b/src/java.base/share/conf/security/java.security
@@ -188,7 +188,7 @@ RestrictedSecurity.NSS.140-2.securerandom.algorithm = PKCS11
 RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.desc.name = OpenJCEPlusFIPS Cryptographic Module FIPS 140-3
 RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.desc.default = false
 RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.desc.fips = true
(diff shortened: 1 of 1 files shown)


JinhangZhang:
@jasonkatonica FYI

src/java.base/share/conf/security/java.security
```diff
@@ -188,7 +188,7 @@ RestrictedSecurity.NSS.140-2.securerandom.algorithm = PKCS11
 RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.desc.name = OpenJCEPlusFIPS Cryptographic Module FIPS 140-3
 RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.desc.default = false
 RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.desc.fips = true
-RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.desc.hash = SHA256:b3e01416cda6ca0134e92940b1c2d94664a742325cbbc8d65b6f536ea3770853
+RestrictedSecurity.OpenJCEPlusFIPS.FIPS140-3.desc.hash = SHA256:dd19c8f8f2578cf400c11b5c7d003684cba5fc4999ac5c55d2a73099f70f9582
```
keithc-ca, line 191: The hash should be `a5bd1607d552ad08e8d3372bceaedb759a68a48ed8d3840abc1992615a814447` (`jce.provider.3` is different as compared with later versions).

keithc-ca (changes requested):

src/java.base/share/conf/security/java.security
JinhangZhang: Updated.

keithc-ca:

keithc-ca (approved):

PR #830, opened 2024-10-28
Replace sprintf with snprintf in java.base

pshipton:
In files:
check_version.c
jvm_dtrace.c
libjvm_db.c
ResolverConfigurationImpl.c

### DIFF
--- a/src/java.base/solaris/native/libjvm_db/libjvm_db.c
+++ b/src/java.base/solaris/native/libjvm_db/libjvm_db.c
@@ -21,6 +21,11 @@
  * questions.
  *
  */
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2024, 2024 All Rights Reserved
+ * ===========================================================================
+ */
 
 #include <stdio.h>
 #include <stdlib.h>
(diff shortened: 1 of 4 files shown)


pshipton:
There are more but they will need to be fixed in later versions first before backporting.

src/java.base/solaris/native/libjvm_db/libjvm_db.c
```diff
@@ -1327,7 +1332,7 @@ int Jget_vframe(jvm_agent_t* J, int vframe_no,
   int32_t   err;
 
   if (vframe_no >= N->vf_cnt) {
-     (void) sprintf(name, "Wrong inlinedMethod%1d()", vframe_no);
+     (void) snprintf(name, size, "Wrong inlinedMethod%1d()", vframe_no);
```
keithc-ca, line 1335: While you're here, please correct the size argument on line 1340:
```c
  err = name_for_methodPtr(J, vf->method, name + 1, size - 1);
```
(see lines 1097 and 1100 for comparison).

keithc-ca (changes requested):

src/java.base/solaris/native/libjvm_db/libjvm_db.c
pshipton: Updated.

keithc-ca:
Jenkins test sanity.functional xlinux,win jdk11

keithc-ca (approved):

PR #836, opened 2024-11-04
Avoid rebuilding native code headers for OpenJCEPlus

KostasTsiounis:
Some `OpenJCEPlus` headers are already compiled as part of a previous target, so there is no reason to recompile them. Instead an additional variable is passed as part of the make command to indicate that this step can be skipped.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/824


KostasTsiounis:
FYI @keithc-ca

keithc-ca (approved):

keithc-ca:
Jenkins compile alinux jdk11

PR #838, opened 2024-11-04
Fix names of algorithms printed in debug trace

KostasTsiounis:
The parameter used to indicate the name of the algorithm, whose native implementation's availability is being checked, was incorrectly set to the provider name, leading to incorrect debug trace messages.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/874


KostasTsiounis:
FYI @keithc-ca

keithc-ca (approved):

PR #840, opened 2024-11-05
Block all properties that can be set by profile

KostasTsiounis:
Instead of just blocking the properties that are set by the loaded profile, block all properties that can potentially be set by any `RestrictedSecurity` profile.

Back-ported by: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/868


KostasTsiounis:
FYI @keithc-ca

keithc-ca:
Jenkins compile zlinux jdk11

PR #841, opened 2024-11-05
Check if OpenSSL is in FIPS mode and use Java for some algorithms

KostasTsiounis:
If the `OpenSSL` library used is in `FIPS` mode, avoid using it for algorithms that are not `FIPS` compliant and revert to the original Java implementation.

Back-ported by: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/867


KostasTsiounis:
FYI @keithc-ca

keithc-ca:
Jenkins compile plinux jdk11

keithc-ca (approved):

PR #843, opened 2024-11-07
Replace sprintf with snprintf in networking

pshipton:
Backport https://github.com/ibmruntimes/openj9-openjdk-jdk17/pull/406


keithc-ca:
The s390x_linux failure appears to be an intermittent CRIU problem, unrelated to this change.

PR #850, opened 2024-12-03
Modify JSR166TestCase to gather more error information

theresa-m:
Update Xdump to catch JUnit exception and AssertionFailedError

Port from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/898


keithc-ca:
Jenkins test sanity.openjdk alinux jdk11

keithc-ca (approved):

PR #851, opened 2024-12-03
Fix profiles with same name beginnings being treated as one

taoliult:
Fix the issue where profiles with the same beginnings of profile name are treated as a single profile.

This is a back-port PR from JDKNext PR: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/887


keithc-ca:
Jenkins compile alinux jdk11

keithc-ca (approved):

PR #854, opened 2024-12-06
Update SunJSSE fully qualified name for test cases

taoliult:
In JDK11, the fully qualified name for SunJSSE provider is `com.sun.net.ssl.internal.ssl.Provider`. Update this for all the restricted security mode test cases. This update only applies to JDK11.

### DIFF
--- a/closed/test/jdk/openj9/internal/security/property-java.security
+++ b/closed/test/jdk/openj9/internal/security/property-java.security
@@ -36,7 +36,7 @@ RestrictedSecurity.TestBase.Version.jce.certpath.disabledAlgorithms =
 RestrictedSecurity.TestBase.Version.jce.legacyAlgorithms =
 RestrictedSecurity.TestBase.Version.jce.provider.1 = sun.security.provider.Sun
 RestrictedSecurity.TestBase.Version.jce.provider.2 = com.sun.crypto.provider.SunJCE
-RestrictedSecurity.TestBase.Version.jce.provider.3 = sun.security.ssl.SunJSSE
+RestrictedSecurity.TestBase.Version.jce.provider.3 = com.sun.net.ssl.internal.ssl.Provider
 
 RestrictedSecurity.TestBase.Version.javax.net.ssl.keyStore = NONE
 RestrictedSecurity.TestBase.Version.securerandom.provider = OpenJCEPlusFIPS
@@ -50,7 +50,7 @@ RestrictedSecurity.TestBase.Version-Extended.tls.disabledAlgorithms =
 RestrictedSecurity.TestBase.Version-Extended.jce.provider.1 = sun.security.provider.Sun
 RestrictedSecurity.TestBase.Version-Extended.jce.provider.2 = sun.security.rsa.SunRsaSign
 RestrictedSecurity.TestBase.Version-Extended.jce.provider.3 = sun.security.ec.SunEC
-RestrictedSecurity.TestBase.Version-Extended.jce.provider.4 = sun.security.ssl.SunJSSE
+RestrictedSecurity.TestBase.Version-Extended.jce.provider.4 = com.sun.net.ssl.internal.ssl.Provider
 RestrictedSecurity.TestBase.Version-Extended.jce.provider.5 = com.sun.crypto.provider.SunJCE
 RestrictedSecurity.TestBase.Version-Extended.jce.provider.6 = sun.security.jgss.SunProvider
 RestrictedSecurity.TestBase.Version-Extended.jce.provider.7 = com.sun.security.sasl.Provider
@@ -76,7 +76,7 @@ RestrictedSecurity.Test-Profile.Base.tls.disabledAlgorithmsWrongTypo =
 
 RestrictedSecurity.Test-Profile.Base.jce.provider.1 = sun.security.provider.Sun
 RestrictedSecurity.Test-Profile.Base.jce.provider.2 = com.sun.crypto.provider.SunJCE
-RestrictedSecurity.Test-Profile.Base.jce.provider.3 = sun.security.ssl.SunJSSE
+RestrictedSecurity.Test-Profile.Base.jce.provider.3 = com.sun.net.ssl.internal.ssl.Provider
 
 RestrictedSecurity.Test-Profile.Base.securerandom.provider = OpenJCEPlusFIPS
 RestrictedSecurity.Test-Profile.Base.securerandom.algorithm = SHA512DRBG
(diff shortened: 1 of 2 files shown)


closed/test/jdk/openj9/internal/security/property-java.security
```diff
@@ -36,7 +36,7 @@ RestrictedSecurity.TestBase.Version.jce.certpath.disabledAlgorithms =
 RestrictedSecurity.TestBase.Version.jce.legacyAlgorithms =
 RestrictedSecurity.TestBase.Version.jce.provider.1 = sun.security.provider.Sun
 RestrictedSecurity.TestBase.Version.jce.provider.2 = com.sun.crypto.provider.SunJCE
-RestrictedSecurity.TestBase.Version.jce.provider.3 = sun.security.ssl.SunJSSE
+RestrictedSecurity.TestBase.Version.jce.provider.3 = com.sun.net.ssl.internal.ssl.Provider
```
keithc-ca, line 39: Isn't a change to `RestrictedSecurity.TestBase.Version.desc.hash` required because of this update?

closed/test/jdk/openj9/internal/security/property-java.security
```diff
@@ -50,7 +50,7 @@ RestrictedSecurity.TestBase.Version-Extended.tls.disabledAlgorithms =
 RestrictedSecurity.TestBase.Version-Extended.jce.provider.1 = sun.security.provider.Sun
 RestrictedSecurity.TestBase.Version-Extended.jce.provider.2 = sun.security.rsa.SunRsaSign
 RestrictedSecurity.TestBase.Version-Extended.jce.provider.3 = sun.security.ec.SunEC
-RestrictedSecurity.TestBase.Version-Extended.jce.provider.4 = sun.security.ssl.SunJSSE
+RestrictedSecurity.TestBase.Version-Extended.jce.provider.4 = com.sun.net.ssl.internal.ssl.Provider
```
keithc-ca, line 53: Again, shouldn't `RestrictedSecurity.Test-Profile.Base.desc.hash` also need to change?

keithc-ca (changes requested):

closed/test/jdk/openj9/internal/security/property-java.security
taoliult: In the `TestProperties.java` file, all profiles except for `Test-Profile-SameStartWithoutVersion` are designed to test error messages. In these profiles, the error message is returned before the hash value is checked, so the hash value is not used.

closed/test/jdk/openj9/internal/security/property-java.security
taoliult: Same as above, in the `TestProperties.java` file, all profiles except for `Test-Profile-SameStartWithoutVersion` are designed to test error messages. In these profiles, the error message is returned before the hash value is checked, so the hash value is not used.

closed/test/jdk/openj9/internal/security/property-java.security
keithc-ca: It would be helpful to have comments to that effect directly in the property files and perhaps specify hashes of all zeros if they are not meaningful.

keithc-ca: Or they could say something like this:
```
RestrictedSecurity.Test-Profile.Base.desc.hash = SHA256:not-checked
```

taoliult: Since we don’t use these hash values, I think it’s fine to just leave them as they are. However, if we do want to make them more readable, like changing them to "`SHA256:not-checked`", I can update this by creating another PR. For two reasons:

- This change "`SHA256:not-checked`" would need to be ported to all JDK versions.

- I can combine this update with the "fixing/improving these comments" which suggested on the JDK8 PR https://github.com/ibmruntimes/openj9-openjdk-jdk8/pull/784, starting from the head stream.

taoliult: Issue created: https://github.ibm.com/runtimes/jit-crypto/issues/615

keithc-ca: Please create an issue in a public repository, perhaps at https://github.com/eclipse-openj9/openj9/issues.

taoliult:
@pshipton Could you help to approve this since Keith is on vacation?

pshipton (approved):

pshipton:
Does this fix any existing OpenJ9 issues?
I assume you want the change in the 0.49 release?

taoliult:
> Does this fix any existing OpenJ9 issues? I assume you want the change in the 0.49 release?

It updates the restricted security mode test cases for JDK 11 to correct the fully qualified name of the SunJSSE provider. This change is not a must for the current 0.49 release, it can be included in the next release if it doesn't make it for 0.49.

pshipton:
> It updates the restricted security mode test cases for JDK 11 to correct the fully qualified name of the SunJSSE provider.

I understand, but are there any issues opened at https://github.com/eclipse-openj9/openj9/issues that describe the failure and will be closed by merging this?

pshipton:
@jasonkatonica maybe you can help with my question. Was the failure exposed by the triage of the jdk11 M2 builds?

jasonkatonica:
Hi @pshipton , 

I personally don't think we need to make this change. This update seems like stylistic change and will not effect how the test runs. 

This update is only to provide clarity that a hash check is not validated / used in a new test @taoliult created to test our restricted security mode.

jasonkatonica:
Hi Peter,

In terms of the comment https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/854#discussion_r1876806771 I don't believe that this update is necessary. This is a cosmetic update to make it clear that the hash is not used ( lots of configuration is not used in the tests ).

This update in general is not required for the M2 builds / .49. We do not currently run these tests as part of our automatons that I am aware of for this release.

pshipton:
It will likely end up in 0.49 just for convenience. It would be harder to keep it out.

closed/test/jdk/openj9/internal/security/property-java.security
keithc-ca: I think the `.hash` property should make that clear, for example:
```
RestrictedSecurity.TestBase.Version.desc.hash = irrelevant
```

taoliult: Discussed within the team, we think that updating the value of `.hash` is unnecessary. Some test cases are designed for testing the error messages, so some properties in those test case profiles may not be used and checked, we keep them in the profile to maintain the integrity of the profile. The values of those properties do not actually matter, we just simply assign default values to them.

We can continue discussing this in the issue: https://github.com/eclipse-openj9/openj9/issues/20815

PR #859, opened 2025-01-15
Consistently load OpenSSL libraries among different platforms

KostasTsiounis:
As part of this pull request, these changes are performed:
- Allow loading specific user-defined OpenSSL library using the `-Djdk.native.openssl.lib` option.
- OpenSSL library loading is consolidated into a single file using ifdefs.
- The platform-specific MD files are no longer needed and are thus deleted.
- Additional traces are added to indicate the attempts and actual OpenSSL library loaded.
- The location of the library loaded is printed.

Moreover, the order of preference for loading a library is updated to follow this order:

1. Explicitly load what was specified via JVM property. Fail if loading fails.
2. Search within the Semeru directories for a bundled version.
3. Search the system for existing libraries and attempt to find the higher version.
4. If all of the previous steps fail, revert to original Java implementation for crypto.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/884

Co-authored by: Paritosh Kumar <paritkum@in.ibm.com>
Co-authored by: Kostas Tsiounis <kostas.tsiounis@ibm.com>


KostasTsiounis:
FYI @keithc-ca

keithc-ca:
Moving to draft state based on https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/884#issuecomment-2596147020.

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
```diff
@@ -1,6 +1,6 @@
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2024 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2025 All Rights Reserved
```
keithc-ca, line 3: I suggest this file should be identical to its counter part from jdk17; I see no reason we should retain the cosmetic differences.

keithc-ca (changes requested):

closed/src/java.base/share/native/libjncrypto/NativeCrypto.c
KostasTsiounis: Updated to be like in 17 with 7ff900a4a6f8382c145f678aab89878e051417c3.

keithc-ca: Please squash.

KostasTsiounis: Done.

keithc-ca:
Jenkins line endings check

keithc-ca:
Jenkins compile aix,osx,plinux jdk11

keithc-ca (approved):

PR #860, opened 2025-01-20
Use generic rather than release specific link in exclude lists

KostasTsiounis:
In the exclude lists for the `OpenJCEPlusFIPS` weak and strict profile, a generic link is provided for each excluded test until it is investigated. Since the solution will become available for all releases, no release specific link is required.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/918


KostasTsiounis:
@JasonFengJ9 Can you help review?

JasonFengJ9 (approved):

PR #868, opened 2025-02-20
Backport 8347740: java/io/File/createTempFile/SpecialTempFile.java failing

pshipton:
Fix the test so it doesn't fail on Windows Server 2025 machines, without waiting to see if OpenJDK will backport the test fix.

So far there is no indication of a backport in
https://bugs.openjdk.org/browse/JDK-8345368
https://bugs.openjdk.org/browse/JDK-8347740

https://github.com/ibmruntimes/openj9-openjdk-jdk17/commit/0ff2c3742e

### DIFF
--- a/test/jdk/java/io/File/createTempFile/SpecialTempFile.java
+++ b/test/jdk/java/io/File/createTempFile/SpecialTempFile.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2013, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2013, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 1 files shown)


keithc-ca:
Same comment here: https://github.com/ibmruntimes/openj9-openjdk-jdk24/pull/26#issuecomment-2672208265.

What about jdk21 and jdk17?

pshipton:
> What about jdk21 and jdk17?

OpenJDK already did these backports.

pshipton:
Updated.

keithc-ca:
Jdk17 and Jdk21 only appear to have the first of the two changes included here. Why do we not want both for those versions?

pshipton:
They have the newer fixes as of earlier this afternoon. Pls update and check again.

keithc-ca:
The change in jdk17 doesn't look the same (or right to me). Should we fix that first?

pshipton:
You'll have to explain. OpenJDK added a fix in jdk17, I don't see why we need to change it unless it's not working.

keithc-ca:
I don't know whether it will work, but jdk17 includes code that was removed in other versions:
```java
        boolean exceptionExpected =
            !(System.getProperty("os.name").matches("^.*[11|2025]$") ||
              new OSVersion(10, 0).compareTo(OSVersion.current()) > 0);
        test("ReservedName", resvPre, resvSuf, exceptionExpected);
```
The final argument to this this call is non-null as compared with the later call (present in jdk17 and other versions)
```java
        test("ReservedName", resvPre, resvSuf, null);
```

pshipton:
Sure, I can fix it now. I expect OpenJDK will notice failures and also do so for the next build.

keithc-ca (approved):

PR #869, opened 2025-02-20
Enforce a more strict FIPS 140-3 JSSE profile definition

jasonkatonica:
The default JSSE provider enables a few services that we would like to disable by default whenever users are making use of the strict 140-3 FIPS profile. Specific services disabled includes the `PKCS12` KeyStore, `MD5andSHA1withRSA` Signature, and SSLContexts of name `DTLSv1.0`, `TLSv1`, and `TLSv1.1`.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/939


keithc-ca:
Please rebase to avoid the implicit merge.

keithc-ca (approved):

PR #873, opened 2025-02-28
Prohibit the insertion of 3rd-party provider

JinhangZhang:
Some applications want to use CRIU and enable the -XX:CRIUSecProvider JVM option. That being said only CRIUSec provider can be used during the pre checkpoint phase of CRIU. However, some non-CRIUSec providers might be inserted, which contradicts the purpose of our -XX:CRIUSecProvider JVM option. Therefore, we added a check in the insertProviderAt function to prevent the addition of non-CRIUSec providers.

### DIFF
--- a/src/java.base/share/classes/java/security/Security.java
+++ b/src/java.base/share/classes/java/security/Security.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2022, 2024 All Rights Reserved
+ * (c) Copyright IBM Corp. 2022, 2025 All Rights Reserved
  * ===========================================================================
  */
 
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/java/security/Security.java
```diff
@@ -348,13 +348,26 @@ public static String getAlgorithmProperty(String algName,
...
+                if ("CRIUSEC".equals(existingProvider.getName())) {
+                    if (criuDebug) {
+                        System.out.println("Trying to insert + " + provider.getName()
+                                + " during the pre-checkpoint which is not allowed.");
+                    }
+                    throw new RuntimeException("Inserting " + provider.getName()
+                            + " during the pre-checkpoint is not allowed");
+                }
+            }
+        }
```
keithc-ca, line 365: Same comment here as in https://github.com/ibmruntimes/openj9-openjdk-jdk21/pull/260#pullrequestreview-2651292078.

keithc-ca (changes requested):

src/java.base/share/classes/java/security/Security.java
JinhangZhang: Moved `ProviderList newList = ProviderList.insertAt(list, provider, position - 1);` up.

PR #875, opened 2025-03-04
Allow multiple constraints for each algorithm

KostasTsiounis:
If a constraint for an algorithm is found and the class attempting to utilize it doesn't match the accepted uses, the algorithm is considered not allowed and loading it does not succeed.

Instead, we want to check all available constraints for an algorithm before deciding if it is allowed to be used by a specific module, package and/or class.

Additional test cases are added to check this functionality.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/954


KostasTsiounis:
FYI @keithc-ca

keithc-ca (approved):

PR #878, opened 2025-03-06
Only check and set FIPS mode in FIPS profiles

KostasTsiounis:
A `RestrictedSecurity` profile doesn't have to be a `FIPS` profile, in which case a `FIPS mode` property is not required and the corresponding system property should not be set.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/957


KostasTsiounis:
FYI @keithc-ca

keithc-ca (approved):

PR #880, opened 2025-03-14
Include System.c natives to make use of System.initProperties

pshipton:
Remove the Java_java_lang_System_mapLibraryName native so it can't conflict with the OpenJ9 native of the same name.

Related to https://github.com/eclipse-openj9/openj9/issues/21189


src/java.base/share/native/libjava/System.c
```diff
@@ -354,7 +360,9 @@ Java_java_lang_System_initProperties(JNIEnv *env, jclass cla, jobject props)
         PUTPROP(props, "sun.desktop", sprops->desktop);
     }
 
+#if 0 /* Exclude, unused. OpenJ9 implements it to assert. */
     ret = JVM_InitProperties(env, props);
+#endif /* Exclude, unused. OpenJ9 implements it to assert. */
```
keithc-ca, line 365: This change is not necessary: the OpenJ9 implementation does nothing for Java 11.

pshipton: Removed it.

keithc-ca:

PR #884, opened 2025-03-25
Avoid checking RestrictedSecurity profile hash during jar verification

KostasTsiounis:
If the process of verifying a jar is started before the `RestrictedSecurity` profile is loaded, the hash calculation is triggered as part of it leading to a nested jar verification and a subsequent error.

To avoid that, the hash calulation of a profile is skipped if triggered by a jar verification process and is performed later in the loading process.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/969


KostasTsiounis:
FYI @keithc-ca

keithc-ca (approved):

PR #886, opened 2025-03-26
Avoid loading general symlinks first

KostasTsiounis:
In a few situations, such as in `RHEL` systems in `FIPS` mode, where `OpenSSL` is in `FIPS` mode as well, loading a general symlink is a security concern and leads to failure, so versioned libraries should be preferred.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/966


KostasTsiounis:
FYI @keithc-ca

keithc-ca:
Jenkins compile xlinux jdk11

keithc-ca (approved):

PR #892, opened 2025-04-08
Rename bundled openssl libraries

jasonkatonica:
OpenSSL libraries that are bundled within the runtime are being renamed to avoid conflicts with system libraries that may have the same name. Additional updates were made to the loading logic associated with native cryptography to make use of the correct bundled version by default.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/980


keithc-ca:

PR #895, opened 2025-04-16
Use PrivilegedAction to use StackWalker with RETAIN_CLASS_REFERENCE option

KostasTsiounis:
If a `SecurityManager` is utilized, permissions are needed to use the `StackWalker` class with the `RETAIN_CLASS_REFERENCE` option, which `RestrictedSecurity` doesn't have. A `PrivilegedAction` is used to allow said action in this case.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk21/pull/280

Fixes: https://github.com/eclipse-openj9/openj9/issues/21614


KostasTsiounis:
FYI @keithc-ca

keithc-ca (approved):

PR #896, opened 2025-04-17
Check RestrictedSecurity profile hash after Providers init

KostasTsiounis:
If the `Providers` class hasn't had a chance to initialize, the required message digest algorithms are not available to check the hash value of the `RestrictedSecurity` profile.

With this change, the check is only allowed after the appropriate initialization has been performed.

Fixes: https://github.com/eclipse-openj9/openj9/issues/21615

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/991

### DIFF
--- a/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
+++ b/closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
@@ -79,6 +79,8 @@ public final class RestrictedSecurity {
 
     private static ProfileParser profileParser;
 
+    private static boolean enableCheckHashes;
+
     private static RestrictedSecurityProperties restricts;
 
     private static final Set<String> unmodifiableProperties = new HashSet<>();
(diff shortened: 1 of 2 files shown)


KostasTsiounis:
FYI @keithc-ca

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
```diff
@@ -192,18 +194,27 @@ private static boolean isJarVerifierInStackTrace() {
      * is calculated and compared to the expected value.
      */
     @SuppressWarnings("removal")
```
keithc-ca, line 196: The annotation belongs on the private method.

keithc-ca (changes requested):

closed/src/java.base/share/classes/openj9/internal/security/RestrictedSecurity.java
KostasTsiounis: Removed.

KostasTsiounis: Removed.

keithc-ca (approved):

PR #897, opened 2025-04-25
Enable CRaC for all Linux platforms

JasonFengJ9:
Enable `CRaC` for all Linux platforms

Cherry-pick
* https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/995


keithc-ca (approved):

JasonFengJ9:
@pshipton can this change be integrated into 0.51 release?

FYI @tajila

keithc-ca:
> integrated into 0.51 release

I assume you are also asking about jdk17 and jdk21 - I don't think it makes sense to include it in jdk11 without later versions.

JasonFengJ9:
>> integrated into 0.51 release

> I assume you are also asking about jdk17 and jdk21 - I don't think it makes sense to include it in jdk11 without later versions.

That's correct. Will created porting PRs as per Pete's request.

PR #912, opened 2025-05-13
Use CXX_VERSION_STRING instead of COMPILER_VERSION_STRING

pshipton:
COMPILER_VERSION_STRING seems an internal variable that shouldn't be used.

Backport https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1010


keithc-ca:
Jenkins compile plinux jdk11

keithc-ca (approved):

PR #913, opened 2025-05-13
Properly check if jdk.native.openssl.lib property is set

KostasTsiounis:
Whether the property has been set at all needs to be checked before checking if it's not empty.

Related to issue https://github.com/eclipse-openj9/openj9/issues/21867

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1011


KostasTsiounis:
FYI @keithc-ca

keithc-ca (approved):

PR #916, opened 2025-05-21
Enable the use of OpenJCEPlusFIPS for HKDF operations

KostasTsiounis:
When in `FIPS` mode, the implementation offered by `OpenJCEPlusFIPS` should be utilized for `HKDF` operations, instead of the default one provided by `Sun`.

More specifically, instead of performing the `extract()` and `expand()` operations in `HKDF.java`, the FIPS-compliant methods offered by `OpenJCEPlusFIPS` should be called.

This change is only applicable in Java 24 and below. In Java 25, the entire `HKDF.java` class, which is used independently and not provided through the JCE framework, is removed entirely. The way `HKDF` functionality is offered is being restructured and a new `KDFSpi` is introduced. Whoever wants to provide said functionality from this point on will need to extend the `KDFSpi` and register their service through a provider.

With this new behaviour, `OpenJCEPlusFIPS` can just provide a service that extends the `KDFSpi` (see issue https://github.com/IBM/OpenJCEPlus/issues/599) and the calls for `HKDF` functionality will be appropriately directed to the FIPS-compliant module without any further changes to the `JCK` classes.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk24/pull/85

### DIFF
--- a/src/java.base/share/classes/sun/security/ssl/HKDF.java
+++ b/src/java.base/share/classes/sun/security/ssl/HKDF.java
@@ -23,15 +23,33 @@
  * questions.
  */
 
+/*
+ * ===========================================================================
+ * (c) Copyright IBM Corp. 2025, 2025 All Rights Reserved
+ * ===========================================================================
+ */
+
 package sun.security.ssl;
 
+/*[IF OPENJCEPLUS_SUPPORT]*/
+import java.lang.reflect.Constructor;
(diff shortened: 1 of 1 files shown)


keithc-ca:
Jenkins compile zlinux jdk11

keithc-ca (approved):

PR #925, opened 2025-05-29
Update KeyDestructionTest for FIPS 140-3

jasonkatonica:
This test is failing since the FIPS 140-3 requires RSA key sizes to be 2048 or higher.

Fixes https://github.com/eclipse-openj9/openj9/issues/21921

### DIFF
--- a/test/jdk/javax/security/auth/Destroyable/KeyDestructionTest.java
+++ b/test/jdk/javax/security/auth/Destroyable/KeyDestructionTest.java
@@ -42,7 +42,7 @@ import javax.security.auth.DestroyFailedException;
 public class KeyDestructionTest {
     public static void main(String[] args) throws Exception {
         String kpgAlgorithm = "RSA";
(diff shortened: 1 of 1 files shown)


test/jdk/javax/security/auth/Destroyable/KeyDestructionTest.java
```diff
@@ -42,7 +42,7 @@
 public class KeyDestructionTest {
     public static void main(String[] args) throws Exception {
         String kpgAlgorithm = "RSA";
-        KeyPair keypair = generateKeyPair(kpgAlgorithm, 1024);
+        KeyPair keypair = generateKeyPair(kpgAlgorithm, 2048);
```
pshipton, line 45: This also changes the non FIPS testing. Is there sufficient coverage for keys smaller than 2048 such that this won't allow any regression?

pshipton (approved):

PR #929, opened 2025-06-17
Update test exclusion list for weak FIPS140-3

JinhangZhang:
This commit updates the test exclude list for FIPS140-3 weak profile.


JinhangZhang:
@JasonFengJ9 FYI

PR #951, opened 2025-08-27
Check if service is allowed in RestrictedSecurity mode

KostasTsiounis:
Checks are relocated to verify that a service is allowed, when said service is directly requested from a provider.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1078

### DIFF
--- a/src/java.base/share/classes/java/security/Provider.java
+++ b/src/java.base/share/classes/java/security/Provider.java
@@ -1161,7 +1161,7 @@ public abstract class Provider extends Properties {
         for (Iterator<Map.Entry<ServiceKey, Service>> t =
                 map.entrySet().iterator(); t.hasNext(); ) {
             Service s = t.next().getValue();
-            if (s.isValid() == false) {
+            if ((s.isValid() == false) || !RestrictedSecurity.canServiceBeRegistered(s)) {
                 t.remove();
             }
         }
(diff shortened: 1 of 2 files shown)


KostasTsiounis:
FYI @keithc-ca

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -1339,13 +1343,18 @@ public synchronized Set<Service> getServices() {
...
-                set.addAll(serviceMap.values());
+                serviceMap.values().forEach(service -> {
+                    if (RestrictedSecurity.isServiceAllowed(service)) {
+                        // If allowed by RestrictedSecurity, add it to set.
+                        set.add(service);
+                    }
+                });
             }
             if (legacyMap != null && !legacyMap.isEmpty()) {
                 set.addAll(legacyMap.values());
```
keithc-ca, line 1354: The same filtering as above should occur here:
```java
                legacyMap.values().forEach(service -> {
                    if (RestrictedSecurity.isServiceAllowed(service)) {
                        // If allowed by RestrictedSecurity, add it to set.
                        set.add(service);
                    }
                });
```

KostasTsiounis: Added.

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -1301,6 +1301,10 @@ public Service getService(String type, String algorithm) {
             }
         }
 
+        if (!RestrictedSecurity.isServiceAllowed(s)) {
```
keithc-ca, line 1304: This is missing a `null` check.

keithc-ca (changes requested):

keithc-ca:
Please update the condition on line 1164 of `Provider` for consistency with jdk8.

keithc-ca:
Please squash.

keithc-ca:

keithc-ca (approved):

PR #961, opened 2025-09-18
Cleanup LUDCL cache invalidation and restoration logic

chrisdennis:
This cleans up the invalidation and restoration of cached 'latest user defined classloader' references when passing between user and platform code during object deserialization.

We also take this opportunity to remove the unnecessary `isClassCachingEnabled` instance variable and replace its usage with the equivalent `classByNameCache != null`.

It is known that this fixes #20916, and it may also fix #9912 but that is as yet unvalidated.

(cherry picked from commit 7ce8262620a26dac48a8548b70f42b160ab71415) (cherry picked from commit 6cd2a964e2230aaafa7592de90760b844bd0db95)

### DIFF
--- a/src/java.base/share/classes/java/io/ObjectInputStream.java
+++ b/src/java.base/share/classes/java/io/ObjectInputStream.java
@@ -24,7 +24,7 @@
  */
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2018, 2022 All Rights Reserved
+ * (c) Copyright IBM Corp. 2018, 2025 All Rights Reserved
  * ===========================================================================
  */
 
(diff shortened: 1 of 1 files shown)


src/java.base/share/classes/java/io/ObjectInputStream.java
```diff
@@ -674,10 +665,10 @@ public Object readUnshared() throws IOException, ClassNotFoundException {
         ClassLoader oldCachedLudcl = null;
         boolean setCached = false;
```
keithc-ca, line 666: While you're here, please remove the trailing space from line 666.

chrisdennis: Done.

keithc-ca (approved):

PR #968, opened 2025-09-24
Update FIPS exclusion list

taoliult:
Some tests listed in the JDK Next exclusion list were not backported to earlier JDK versions, likely because the tests themselves had not been backported at that time.

Now, these tests need to be added to the exclusion lists for earlier JDK versions because:

1. The tests have since been backported, or
3. They are not yet backported, but once they are, they will inevitably fail since the required algorithms are not supported in FIPS mode.


taoliult:
PR for release branch: https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/970

keithc-ca:
Please correct the commit message: This is not for the 0.56.0 release.

PR #969, opened 2025-09-25
Remove categories and sort FIPS 140-3 excludes

jasonkatonica:
This update eliminates test bucket identifiers like sanity.openjdk and applies sorting to all entries within the FIPS 140-3 exclusion files. Since exclusion categories are not utilized on a per-test-bucket basis and the files are large, this change streamlines file maintenance and improves readability.

Backported-from: https://github.com/ibmruntimes/openj9-openjdk-jdk11/pull/967


keithc-ca:
If #944 should be included here, `KeytoolOpensslInteropTest` should also be removed from `ProblemList-FIPS140_2.txt`.

jasonkatonica:
OK missed that file. Removed `KeytoolOpensslInteropTest` from the 140-2 exclusion file.

PR #1064, opened 2026-08-14
Allow aliases in RestrictedSecurity constraints

KostasTsiounis:
When defining constraints for specific services allowed by specific providers, one can now use an alias instead of the original algorithm name.

A test case has, also, been added to verify said functionality.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1241

---------
- [x] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).

### DIFF
--- a/src/java.base/share/classes/java/security/Provider.java
+++ b/src/java.base/share/classes/java/security/Provider.java
@@ -25,7 +25,7 @@
 
 /*
  * ===========================================================================
- * (c) Copyright IBM Corp. 2022, 2025 All Rights Reserved
+ * (c) Copyright IBM Corp. 2022, 2026 All Rights Reserved
  * ===========================================================================
  */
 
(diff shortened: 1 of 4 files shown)


KostasTsiounis:
FYI @keithc-ca

keithc-ca:
Jenkins compile xlinux jdk11

src/java.base/share/classes/java/security/Provider.java
```diff
@@ -1405,7 +1405,7 @@ protected void putService(Service s) {
             throw new IllegalArgumentException
                     ("service.getProvider() must match this Provider object");
         }
-        if (!RestrictedSecurity.canServiceBeRegistered(s)) {
+        if (!RestrictedSecurity.canServiceBeRegistered(s, s.getAliases())) {
```
keithc-ca, line 1408: The use of `canServiceBeRegistered()` on line 1164 must be updated as well.

keithc-ca (changes requested):

keithc-ca:
Please fix the compile error.

KostasTsiounis:
> Please fix the compile error.

Done.

src/java.base/share/classes/java/security/Provider.java
KostasTsiounis: Added.

keithc-ca:
Jenkins compile alinux jdk11

keithc-ca:
The new tests passed in https://openj9-jenkins.osuosl.org/job/Grinder/5346.

keithc-ca (approved):

PR #1072, opened 2026-09-08
Fix regex to accept wildcard character for constraint algorithm

KostasTsiounis:
One should be able to use a wildcard `*` character for the algorithm part of a `RestrictedSecurity` provider constraint. The regex parsing that portion was wrong and missing the allowance for the character. This is fixed now.

Back-ported from: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1262

---------
- [X] I confirm that I make this contribution in accordance with the [OpenJDK Interim AI Policy](https://openjdk.org/legal/ai).


KostasTsiounis:
FYI @keithc-ca

keithc-ca (approved):

PR #1075, opened 2026-09-14
Update CheckSecurityProvider and UnsupportedProvider tests

taoliult:
Update CheckSecurityProvider and UnsupportedProvider tests to account for the new OpenJCEPlusSemeruDefaults provider added to java.security.

- CheckSecurityProvider.java: add the expected class name "com.ibm.crypto.plus.provider.OpenJCEPlusSemeruDefaults" before SUN in the ordered provider list, matching the position defined in java.security.

- UnsupportedProvider.java: exclude providers whose name starts with "OpenJCEPlusSemeruDefaults" from the SHA-3 supported set, since the provider does not support SHA-3 digest algorithms. (This test only exists in JDK 11, so the change only needs to be applied to the JDK 11 repository.)

This is a back port PR from PR: https://github.com/ibmruntimes/openj9-openjdk-jdk/pull/1272


test/jdk/java/lang/SecurityManager/CheckSecurityProvider.java
```diff
@@ -70,6 +70,8 @@ public static void main(String[] args) throws Exception {
             layer.findModule("jdk.crypto.cryptoki")
                 .ifPresent(m -> expected.add("sun.security.pkcs11.SunPKCS11"));
         }
+        layer.findModule("openjceplus")
```
keithc-ca, line 73: While you're here, please fix the typo on line 60 ("the the" -> "the")?

test/jdk/java/security/MessageDigest/UnsupportedProvider.java
```diff
@@ -70,7 +70,8 @@ public static void main(String args[]) {
 
     // Check if specific provider supports SHA-3 hash algorithms
     static boolean isSHA3Supported(String provName) {
-        if ("SUN".equals(provName) || provName.startsWith("OpenJCEPlus")) {
+        if ("SUN".equals(provName) || (provName.startsWith("OpenJCEPlus")
+                && !provName.startsWith("OpenJCEPlusSemeruDefaults"))) {
```
keithc-ca: The copyright date in this file should be updated.

Suggest folding this to clarify the structure of the condition:
```java
        if ("SUN".equals(provName)
                || (provName.startsWith("OpenJCEPlus")
                        && !provName.startsWith("OpenJCEPlusSemeruDefaults"))
        ) {
```

keithc-ca (changes requested):

test/jdk/java/lang/SecurityManager/CheckSecurityProvider.java
taoliult: Updated.

test/jdk/java/security/MessageDigest/UnsupportedProvider.java
taoliult: Updated.

keithc-ca (approved):

